PR #27, opened 2021-03-30
Clear up

yz89:
1. remove genesis TotalBorrow and TotalSupply
2. remove staking call from loans pallet


GopherJ:
I think it looks good

PR #28, opened 2021-03-30
Cumulus

GopherJ:
- [x] use rococo-v1 dependencies, cumulus, substrate, polkadot-xcm and make it compiles
- [x] integrate our pallets
- [x] fix mock & tests
- [x] update README.md with parachain guide & parallel info
- [x] add stricter CI to enforce styles & tests & lint & wasm build
- [x] fix current CI
- [x] rabase latest changes of master
- [x] squash commits
- [x] add parallelfinance info(homepage etc)
- [ ] fix runtime cargo check & wasm check


node/Cargo.toml
```diff
@@ -1,61 +1,78 @@
...
+polkadot-service      = { git = 'https://github.com/paritytech/polkadot.git', branch = 'rococo-v1' }
+polkadot-test-service = { git = 'https://github.com/paritytech/polkadot.git', branch = 'rococo-v1' }
+
+[build-dependencies]
+substrate-build-script-utils = { version = '3.0.0', git = 'https://github.com/paritytech/substrate.git', branch = 'rococo-v1' }
 
 [[bin]]
 name = 'parallel'
 
+[package]
```
kaichaosun, line 63: This section should be put at the beginning.

resources/para-200-genesis
```diff
@@ -0,0 +1 @@
+0x0000000000000000000000000000000000000000000000000000000000000000006eceeb89f3ba476aff716ddb165a3aa307dcb0c5fa6348d067990e446b55f7f403170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c11131400
```
kaichaosun, line 1: This file and wasm file should not in the git repo.

runtime/src/lib.rs
```diff
@@ -92,38 +66,33 @@ pub mod opaque {
...
 
-/// This determines the average expected block time that we are targeting.
+/// This determines the average expected block time that we are targetting.
 /// Blocks will be produced at a minimum duration defined by `SLOT_DURATION`.
 /// `SLOT_DURATION` is picked up by `pallet_timestamp` which is in turn picked
 /// up by `pallet_aura` to implement `fn slot_duration()`.
 ///
 /// Change this to adjust the block time.
-pub const MILLISECS_PER_BLOCK: u64 = 6000;
+pub const MILLISECS_PER_BLOCK: u64 = 12000;
```
kaichaosun, line 95: Can you share more about this change?

runtime/src/lib.rs
```diff
@@ -368,19 +266,15 @@ where
 
         let raw_payload = SignedPayload::new(call, extra)
             .map_err(|e| {
-                debug::warn!("SignedPayload error: {:?}", e);
+                log::warn!("SignedPayload error: {:?}", e);
```
kaichaosun, line 269: If error happens, can not make the log level to error?

runtime/src/lib.rs
GopherJ: @kaichaosun from my understanding, rococo doesn't support 6000 block producing time. as gavin wood said during beijing meetings, now rococo only supports 12 seconds.

resources/para-200-genesis
GopherJ: while initializing a prarachain on polkadot.js we need this `genesis` data and the `wasm` file, I saw that some others repo are keeping them as well.

yz89:
We can look at Acala Mandala TC, which is running in development mode.  They use the `alaca-dev' cli and its runtime to achieve this.

https://github.com/AcalaNetwork/Acala/tree/master/bin/acala-dev

GopherJ:
> We can look at Acala Mandala TC, which is running in development mode. They use the `alaca-dev` cli and its runtime to achieve this.
> https://github.com/AcalaNetwork/Acala/tree/master/bin/acala-dev
> https://github.com/AcalaNetwork/Acala/blob/master/Makefile#L3
> 
> Maybe we can merge this pr firstly. Any suggestion about this? @kaichaosun

I tried this but it doesn't produce block yet

MrPai (commented):
I can not compile successfully on a remote ubuntu system, neither the nightly nor stable toolchain. 
the error is "could not compile `polkadot-service`"

yz89 (approved):

kaichaosun (approved):
👍

PR #44, opened 2021-04-09
use USDT instead of USDC

yz89:


GopherJ (approved):

MrPai (approved):
As I reviewed, only change USDC into USDT, I think we can merge it directly.

Issue #32, opened 2021-04-05
Refactor code in loans pallet, mainly storages and dispatchables


kaichaosun:
Items needed to be refactored includes,
- comments/rustdocs for all errors/events, dispatchables, storages
- minimum storage usage, bounded execution/storage, short function
- add traits for pallet to abstract the behaviour
- move the main pallet implementation in lib.rs
- extract to new pallets if storages/dispatchables can be separated
- exchange_rate to use fixed point number, rational of precision
- pub mod module change to pallet, put impl Pallet out of this module
- license

yz89:
comments/rustdocs for all errors/events, dispatchables, storages

Issue #31, opened 2021-04-05
Test coverage for dispatchables in loans pallet


kaichaosun:
This is the list of all existing dispatchables:
- mint
- redeem
- redeem_all
- borrow
- repay_borrow
- repay_borrow_all
- transfer_token
- collateral_asset
- liquidate_borrow

yz89:
Done

PR #51, opened 2021-04-14
Test coverage for dispatchables in loans pallet #31

yz89:
1. Add essential unit test for loans pallet #31
2. Add some rust document for loans pallet #32


pallets/loans/src/mock.rs
```diff
@@ -129,17 +132,37 @@ impl pallet_balances::Config for Runtime {
     type WeightInfo = pallet_balances::weights::SubstrateWeight<Runtime>;
 }
 
+lazy_static! {
+    pub static ref PRICE_ORACLE: Mutex<HashMap<CurrencyId, Option<PriceDetail>>> = {
+        Mutex::new(
+            vec![DOT, KSM, BTC, USDT, XDOT]
```
kaichaosun, line 138: BTC will not be supported until further discussion, we should remove it from code and test cases.

pallets/loans/src/mock.rs
```diff
@@ -129,17 +132,37 @@ impl pallet_balances::Config for Runtime {
...
+                .iter()
+                .map(|&x| (x, Some((1, 1))))
+                .collect(),
+        )
+    };
+}
+
+impl PRICE_ORACLE {
```
kaichaosun, line 146: How about `MockPriceFeeder`?

primitives/src/lib.rs
```diff
@@ -87,7 +87,7 @@ pub type BlockId = generic::BlockId<Block>;
 pub use sp_runtime::OpaqueExtrinsic as UncheckedExtrinsic;
 
 #[derive(Encode, Decode, Eq, PartialEq, Copy, Clone, RuntimeDebug, PartialOrd, Ord)]
-#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize, Hash))]
```
kaichaosun, line 90: Can you explain about the change here?

kaichaosun (approved):
Looks good 👍

pallets/loans/src/mock.rs
yz89: Good idea, I will do that.

primitives/src/lib.rs
yz89: The trait `std::hash::Hash` should be satisfied by CurrencyId enum when use it as the key of  `std::collections::HashMap`.

pallets/loans/src/mock.rs
yz89: All the currency types in the lending market will be listed by the governance module. Current markets are more of a demonstration.

pallets/loans/src/mock.rs
yz89: I will rename it to `MOCK_PRICE_FEEDER` because the static variable should have an upper case name.

MrPai (approved):
looks good!

PR #55, opened 2021-04-16
use nightly as default

yz89:


kaichaosun (commented):
The [docs](https://substrate.dev/docs/en/knowledgebase/getting-started/#2-rust-developer-environment) in substrate.dev shows to use stable as default, we better stick with it. It should be raised as issue by the community if stable is not working for most users.

yz89:
Ok, got it.

PR #56, opened 2021-04-16
fix unit test for ci

yz89:
The static `MOCK_PRICE_FEEDER` stores prices globally that will affect different unit test contexts. 
It must be reset at the end of the test, which changed the price.

### DIFF
--- a/pallets/loans/src/mock.rs
+++ b/pallets/loans/src/mock.rs
@@ -150,6 +150,12 @@ impl MOCK_PRICE_FEEDER {
             .unwrap()
             .insert(currency_id, Some((price, 1u64)));
     }
+
+    pub fn reset() {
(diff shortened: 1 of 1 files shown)


pallets/loans/src/mock.rs
```diff
@@ -150,6 +150,11 @@ impl MOCK_PRICE_FEEDER {
             .unwrap()
             .insert(currency_id, Some((price, 1u64)));
     }
+    pub fn reset() {
```
kaichaosun, line 153: Add an empty line before the function.

kaichaosun (approved):
Looks good, one small concern.

pallets/loans/src/mock.rs
yz89: Ok

PR #64, opened 2021-04-21
use fixed-point number for exchange_rate and multiplier

yz89:
- define type Rate, Ratio, Multiplier
- use the fixed-point number to simplify floating-point arithmetic #32

### DIFF
--- a/pallets/loans/src/loan.rs
+++ b/pallets/loans/src/loan.rs
@@ -76,15 +76,15 @@ impl<T: Config> Pallet<T> {
         mint_amount: Balance,
     ) -> DispatchResult {
         let exchange_rate = Self::exchange_rate(currency_id);
-        let collateral =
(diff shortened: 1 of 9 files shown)


pallets/loans/src/loan.rs
```diff
@@ -207,9 +205,8 @@ impl<T: Config> Pallet<T> {
             return Err(Error::<T>::OracleCurrencyPriceNotReady);
         }
 
-        (collateral_factor * collateral)
-            .checked_mul(exchange_rate)
-            .and_then(|r| r.checked_div(RATE_DECIMAL))
+        exchange_rate
+            .checked_mul_int(collateral_factor * collateral)
```
MrPai, line 209: looks great, 
I have some concerns here, seems like it will return a fixed-point type, but this method will return the Balance type, if this is ok here?

PR #73, opened 2021-04-25
Add orml oracle and prices pallet

RainFallsSilent:


pallets/prices/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+pub mod weights;
+
+pub use module::*;
+pub use weights::WeightInfo;
+
+pub const CURRENCY_DECIMAL: u32 = 18;
+
+pub type OraclePrice = FixedU128;
+
+pub trait PriceProvider<CurrencyId> {
```
MrPai, line 61: Good Job, I think we may need to discuss here, since we already have a trait "[PriceFeeder](https://github.com/parallel-finance/parallel/blob/master/primitives/src/lib.rs#L112)", we may need one of them

MrPai: actually in other pallets, such as pallet loans, will use "PriceFeeder" to get the price, it seems like now we need the trait "PriceProvider" to get the price, so pallet loans need to change their code too.
maybe we can simply implement one single method just like "PriceFeeder" do, and by annotating other methods, 
we could hear others views

RainFallsSilent: Is it better to add a "PriceFeeder" implementation while keeping the "PriceProvider" implementation?

node/src/chain_spec.rs
```diff
@@ -180,6 +181,10 @@ fn development_genesis(
                 .collect(),
         },
         pallet_sudo: vanilla_runtime::SudoConfig { key: root_key },
+		orml_oracle_Instance2: VanillaOracleConfig {
```
yz89, line 184: orml_oracle_Instance2 and orml_oracle_Instance1 are not reasonable names.

pallets/prices/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "pallet-prices"
+authors = ['Parallel Team']
+edition = '2018'
+version = '3.0.0'
```
GopherJ, line 5: let's change version to 1.0.0 like in #80

runtime/vanilla/src/lib.rs
```diff
@@ -374,6 +375,50 @@ where
...
+impl DataFeeder<CurrencyId, OraclePrice, AccountId> for AggregatedDataProvider {
+    fn feed_value(_: AccountId, _: CurrencyId, _: OraclePrice) -> DispatchResult {
+        Err("Not supported".into())
+    }
+}
+
+impl pallet_prices::Config for Runtime {
+    type Event = Event;
+    type Source = AggregatedDataProvider;
+	type GetStableCurrencyId = GetStableCurrencyId;
```
GopherJ, line 416: this should be aligned

runtime/vanilla/src/lib.rs
```diff
@@ -396,7 +441,9 @@ construct_runtime!(
         Loans: pallet_loans::{Pallet, Call, Storage, Event<T>, Config},
         Staking: pallet_staking::{Pallet, Call, Storage, Event<T>, Config},
         OcwOracle: pallet_ocw_oracle::{Pallet, Call, Storage, Event<T>, ValidateUnsigned},
+        VanillaOracle: orml_oracle::<Instance1>::{Pallet, Storage, Call, Config<T>, Event<T>},
```
GopherJ, line 444: do we need two oracles?

OcwOracle, VanillaOracle

pallets/prices/src/lib.rs
RainFallsSilent: Fixed: use "PriceProvider" only.

node/src/chain_spec.rs
RainFallsSilent: Fixed: use orml_oracle_Instance1 only.

runtime/vanilla/src/lib.rs
RainFallsSilent: Fixed

pallets/prices/Cargo.toml
RainFallsSilent: Fixed

runtime/vanilla/src/lib.rs
RainFallsSilent: Ok, will remove OcwOracle later

MrPai:
it seems like there is no API for frontend,

pallets/prices/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// Get price and timestamp by currency_id
+    /// Timestamp is zero and the currency is not stable currency means the price is emergency price
+    fn get_price(currency_id: &CurrencyId) -> Option<PriceDetail> {
+        // if is stable currency, return fixed price
+        if *currency_id == T::GetStableCurrencyId::get() {
+            Some((T::StableCurrencyFixedPrice::get().into_inner(), 0))
+        } else {
+            // if emergency price exists, return it, otherwise return latest price from oracle.
+            Self::get_emergency_price(currency_id).or_else(|| {
+                T::Source::get_no_op(&currency_id)
```
MrPai, line 133: as I check, `get_no_op` won't store the combined data, so we need to combine t every time, I am not sure if this is ok?

yz89 (approved):

pallets/prices/src/lib.rs
RainFallsSilent: Fixed

PR #86, opened 2021-05-01
remove stable currency

RainFallsSilent:


runtime/vanilla/src/lib.rs
```diff
@@ -398,15 +392,18 @@ pub type TimeStampedPrice = orml_oracle::TimestampedValue<OraclePrice, Moment>;
 pub struct AggregatedDataProvider;
 impl DataProvider<CurrencyId, TimeStampedPrice> for AggregatedDataProvider {
     fn get(key: &CurrencyId) -> Option<TimeStampedPrice> {
-        VanillaOracle::get(key)
+        if let Some(price) = VanillaOracle::get(key) {
+            log::info!("DataProvider get price {:?}", price);
```
yz89, line 396: I think you can print Option<Price> directly.

yz89 (approved):

runtime/vanilla/src/lib.rs
RainFallsSilent: fixed:  removed the log directly.

PR #90, opened 2021-05-03
change get_price to return price with decimal 1_000

RainFallsSilent:


pallets/prices/src/lib.rs
```diff
@@ -116,8 +116,14 @@ impl<T: Config> PriceFeeder for Pallet<T> {
     fn get_price(currency_id: &CurrencyId) -> Option<PriceDetail> {
         // if emergency price exists, return it, otherwise return latest price from oracle.
         Self::get_emergency_price(currency_id).or_else(|| {
-            T::Source::get(&currency_id)
-                .and_then(|price| Some((price.value.into_inner(), price.timestamp)))
+            T::Source::get(&currency_id).and_then(|price| {
+                price
+                    .value
+                    .into_inner()
+                    .checked_mul(TOKEN_DECIMAL)
```
MrPai, line 123: It seems like `TOKEN_DECIMAL` is 10^18, and `price.value.into_inner()` is also 10^18,
if this is ok?

RainFallsSilent: fixed

yz89 (approved):

PR #96, opened 2021-05-09
Partial staking onchain logic

kaichaosun:


pallets/staking/src/lib.rs
```diff
@@ -12,86 +12,149 @@
...
             Balance = Balance,
             Amount = Amount,
         >;
 
-        /// The loan's module id, keep all collaterals of CDPs.
+        /// Currency used for staking
+        #[pallet::constant]
+        type StakingCurrency: Get<CurrencyId>;
+
+        /// Currency used for liquid voucher
```
GopherJ, line 68: I think voucher isn't the same thing with staking mecanism as voucher means `money replacement`, people can spend it.

However, staking just ensure that people can get money back, so I think it's not the same thing

pallets/staking/src/lib.rs
```diff
@@ -111,53 +174,157 @@ pub mod module {
...
+        /// May only be called from `T::WithdrawOrigin`.
+        ///
+        /// - `agent`: the multisig account of relay chain.
+        /// - `amount`: the requested assets.
         #[pallet::weight(10_000)]
         #[transactional]
-        pub fn unstake(origin: OriginFor<T>, amount: Balance) -> DispatchResultWithPostInfo {
-            let who = ensure_signed(origin)?;
-            Self::unstake_internal(&who, amount)?;
+        pub fn withdraw(
```
GopherJ, line 230: it looks like that people can stake 1 dot then withdraw 100dot as there is no limit?

pallets/staking/src/lib.rs
```diff
@@ -111,53 +174,157 @@ pub mod module {
...
+            agent: T::AccountId,
+            amount: Balance,
+        ) -> DispatchResultWithPostInfo {
+            T::WithdrawOrigin::ensure_origin(origin)?;
+            ensure!(
+                amount <= T::MaxWithdrawAmount::get(),
+                Error::<T>::ExcessWithdraw
+            );
+
+            T::Currency::transfer(
```
GopherJ, line 241: will this influence exchange rate?  As total staking has changed

pallets/staking/src/lib.rs
kaichaosun: This function can only be called by a specific collective, the collective may be multisig accounts in parachain, which also have a corresponding account in relay chain.

kaichaosun, line 230: A better name.

pallets/staking/src/lib.rs
```diff
@@ -111,53 +174,157 @@ pub mod module {
...
+        /// Record the staking rewards, no real transfer.
+        /// TODO restrict the times an account can report in one day and max rewards.
+        ///
+        /// May only be called from `T::WithdrawOrigin`.
+        ///
+        /// - `amount`: the rewarded assets.
         #[pallet::weight(10_000)]
         #[transactional]
-        pub fn return_pending_balance(
+        pub fn record_rewards(
```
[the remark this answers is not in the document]
kaichaosun: sudo remove rewards operation.

yz89 (approved):

PR #112, opened 2021-05-13
Add multisig pallet and price query RPC

RainFallsSilent:


node/parallel/src/rpc.rs
```diff
@@ -0,0 +1,63 @@
+#![warn(missing_docs)]
```
GopherJ, line 1: I think we should modify https://github.com/parallel-finance/parallel/blob/master/node/parallel/src/service.rs#L323 instead

node/parallel/src/service.rs
```diff
@@ -161,7 +161,7 @@ async fn start_node_impl<RB>(
     polkadot_config: Configuration,
     id: ParaId,
     validator: bool,
-    rpc_ext_builder: RB,
+    _rpc_ext_builder: RB,
```
GopherJ, line 164: IMHO, we should create a `rpc_ext_builder` instead of  ignoring it

GopherJ:
BTW @RainFallsSilent you can add `close: <Issue Id>` to link the target issue, so that we this PR gets merged, the referenced issue can be closed automatically

node/parallel/src/service.rs
RainFallsSilent: rpc_ext_builder depends on transaction_pool which will be created in the start_node_impl function, so I removed this parameter instead.

node/parallel/src/rpc.rs
RainFallsSilent: Fixed

RainFallsSilent:
oracle_getValue:
```json
curl http://localhost:9933 -H "Content-Type:application/json;charset=utf-8" -d   '{"jsonrpc":"2.0", "id":1, "method":"oracle_getValue", "params": ["Aggregated", "DOT"]}'
{"jsonrpc":"2.0","result":{"timestamp":1620871056003,"value":"1111"},"id":1}
```
oracle_getAllValues:
```json
curl http://localhost:9933 -H "Content-Type:application/json;charset=utf-8" -d   '{"jsonrpc":"2.0", "id":1, "method":"oracle_getAllValues", "params": ["Aggregated"]}'   
{"jsonrpc":"2.0","result":[["DOT",{"timestamp":1620843342002,"value":"137000000000000000000"}]],"id":1}
```

GopherJ (approved):

node/parallel/src/service.rs
```diff
@@ -155,24 +155,16 @@ pub fn new_partial(
 ///
 /// This is the actual implementation that is abstract over the executor and the runtime api.
 #[sc_tracing::logging::prefix_logs_with("Parachain")]
-async fn start_node_impl<RB>(
```
MrPai, line 158: if it's ok to change this method signature, could you introduce more about it?

node/parallel/src/service.rs
```diff
@@ -155,24 +155,16 @@ pub fn new_partial(
...
+async fn start_node_impl(
```
MrPai, line 158: if it's ok to change this method signature, could you introduce more details about it?

RainFallsSilent: Do you mean if it's ok to remove the 'rpc_ext_builder' parameter of start_node_impl?
I think it is only a choice to create rpc_ext_builder before or in the start_node_impl function.
But the creation of rpc_ext_builder depends on the transaction pool which is created in the start_node_impl function, so removed the parameter of rpc_ext_builder directly.

GopherJ:
do you think we can merge ? @kaichaosun

PR #120, opened 2021-05-17
Remainings of staking on chain logic

kaichaosun:


pallets/staking/src/lib.rs
```diff
@@ -279,6 +299,37 @@ pub mod pallet {
...
+        ///
+        /// May only be called from `T::WithdrawOrigin`.
+        ///
+        /// - `agent`: the multisig account of relay chain.
+        /// - `amount`: the rewarded assets.
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn record_slash(
+            origin: OriginFor<T>,
+            agent: T::AccountId,
```
ayushmishra2005, line 313: Can we remove this? We are not using this except sending in event.

pallets/staking/src/lib.rs
```diff
@@ -327,6 +378,110 @@ pub mod pallet {
...
+        ///
+        /// - `agent`: the multisig account of relay chain.
+        /// - `owner`: the account which performs `unstake` operation
+        /// - `amount`: the assets can be unbond for the owner's unstaking request.
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn process_pending_unstake(
+            origin: OriginFor<T>,
+            agent: T::AccountId,
+            owner: T::AccountId,
```
ayushmishra2005, line 394: Is the owner different from origin?

pallets/staking/src/lib.rs
```diff
@@ -327,6 +378,110 @@ pub mod pallet {
...
+        ///
+        /// - `agent`: the multisig account of relay chain.
+        /// - `owner`: the account which performs `unstake` operation
+        /// - `amount`: the assets already unbond for the owner's unstaking request.
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn finish_processed_unstake(
+            origin: OriginFor<T>,
+            agent: T::AccountId,
+            owner: T::AccountId,
```
ayushmishra2005, line 446: Same as above. Can we use origin as owner?

pallets/staking/src/lib.rs
kaichaosun: If no agent, there will no info to track which agent introduce slash, now we use the event to keep this info, but can be stored in storage in the future if need.

pallets/staking/src/lib.rs
kaichaosun: The origin should be the one to one map to agent, need also to be multisig.

MrPai (commented):
looks good for now, I'm thinking about what if we send a msg twice?
because `stake-client` acts as an oracle that feeds data to para-chain, we should consider data duplication. I build an issue to track this. [121](https://github.com/parallel-finance/parallel/issues/121)

ayushmishra2005 (approved):

Issue #35, opened 2021-04-05
Test coverage and refactor liquidate pallet


kaichaosun:
Potencial tasks:
- test coverage for execute_liquidation
- short liquidate function by extracting a few small functions
- remove outer loop if possible
- more liquidation workflow, like give general user chance to process liquidation
- comments/rustdocs
- benchmark (split to another issue if needed)
- license

PR #140, opened 2021-05-24
Liquidation refactor

kaichaosun:
The old liquidate pallet will remove later.
The liquidation using offchain worker still not sufficient for production, a client can be used by any community members should be much better.


pallets/liquidate-new/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+                }
+
+                let mut new_loans_detail = loans_detail.clone();
+                new_loans_detail.sort_by(|a, b| a.value.cmp(&b.value));
+                let liquidate_loans = &new_loans_detail[0];
+
+                if let Some(item) = collateral.1.iter().find(|collateral_item| {
+                    collateral_item.value
+                        >= T::LiquidateFactor::get().mul_floor(liquidate_loans.value)
+                }) {
```
MrPai, line 268: Looks perfect. 
Here I'm thinking what if we have multi collateral currencies(such as DOT/KSM/BTC), but we only borrow USDT out, so the loan will only be USDT, in this situation, no single collateral's value will be bigger than the loans.
please fix me if I have any misunderstanding.

MrPai (commented):
Great work.

pallets/liquidate-new/src/lib.rs
MrPai: #35

kaichaosun: You are correct. Use offchain worker to deal this scenario is not flexible, we need a better solution to backup/replace this one.

yz89:
Nice work!

PR #144, opened 2021-05-25
change currency value from u128 to FixedU128

RainFallsSilent:


GopherJ (approved):

GopherJ:
some tests are broken @RainFallsSilent

PR #153, opened 2021-05-27
Rename liquid staking and liquidation pallet

kaichaosun:


node/parallel-dev/src/chain_spec.rs
```diff
@@ -253,7 +253,7 @@ fn testnet_genesis(
                 (CurrencyId::xDOT, Ratio::from_percent(15)),
             ],
         },
-        pallet_staking: vanilla_runtime::StakingConfig {
+        pallet_liquid_staking: vanilla_runtime::LiquidStakingConfig {
```
MrPai, line 256: After we rename the pallet, `stake-client` should change the name too, let's keep it in mind.

GopherJ (approved):

PR #170, opened 2021-05-29
use with_transaction when accruing interest

yz89:
close: #161

### DIFF
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -18,7 +18,12 @@
 
 pub use crate::rate::{InterestRateModel, APR};
 use crate::util::*;
-use frame_support::{log, pallet_prelude::*, transactional, PalletId};
+use frame_support::{
+    log,
+    pallet_prelude::*,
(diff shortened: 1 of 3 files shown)


GopherJ (approved):

pallets/loans/src/lib.rs
```diff
@@ -428,18 +433,20 @@ pub mod module {
...
-                    // This should never happen...
-                    log::info!(
-                        "Could not initialize block!!! {:#?} {:#?}",
-                        block_number,
-                        err
-                    );
-                    0
+            with_transaction(|| {
+                match <Pallet<T>>::accrue_interest() {
+                    Ok(()) => TransactionOutcome::Commit(1000),
```
GopherJ, line 438: what does 1000 mean ?

yz89: It should be the total weight consumed in the `accrue_interest`. Consider that there is no benchmark build for `accrue_interest`, I use 1000 for now.

PR #171, opened 2021-05-29
enable benchmarks in dockerimage & allow reserve_origin to add/reduce…

GopherJ:
… at anytime

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -6,7 +6,7 @@ WORKDIR /parallel
 
 COPY . /parallel
 
-RUN cargo build --$PROFILE --bin parallel-dev
+RUN cargo build --$PROFILE --bin parallel-dev --features runtime-benchmarks
 
 # ===== SECOND STAGE ======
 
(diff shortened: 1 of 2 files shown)


yz89:
why do you want to allow reserve_origin to add/reduce any kinds of currencies?

GopherJ:
> why do you want to allow reserve_origin to add/reduce any kinds of currencies?

because sometimes we may shutdown a market but needs to do operations on reserve I think. Will that be possible?

Dockerfile.release
```diff
@@ -6,7 +6,7 @@ WORKDIR /parallel
 
 COPY . /parallel
 
-RUN cargo build --$PROFILE --bin parallel-dev
+RUN cargo build --$PROFILE --bin parallel-dev --features runtime-benchmarks
```
kaichaosun, line 9: This feature should only be used in benchmark, not the online binary.

GopherJ: it'll make benchmark on our server easier I think

PR #176, opened 2021-05-29
modify to feed price with decimal

RainFallsSilent:


pallets/prices/src/lib.rs
```diff
@@ -108,14 +115,22 @@ impl<T: Config> PriceFeeder for Pallet<T> {
...
         // if emergency price exists, return it, otherwise return latest price from oracle.
-        Self::get_emergency_price(currency_id)
-            .or_else(|| T::Source::get(&currency_id).map(|price| (price.value, price.timestamp)))
+        Self::get_emergency_price(currency_id).or_else(|| {
+            T::Source::get(&currency_id).and_then(|p| {
+                10u128.checked_pow(p.value.decimal.into()).and_then(|d| {
+                    p.value
+                        .price
+                        .checked_div(&FixedU128::from_inner(d))
+                        .and_then(|price| Some((price, p.timestamp)))
```
kaichaosun, line 124: This will lose precision?

RainFallsSilent: FixedU128::from_inner(10_000_000_000) = FixedU128(0.00000001),
Lose precision will only happen when the decimal is larger than 18, and lose precision will begin from the last decimal. I think current currencies will have no precision loss.

yz89 (approved):

PR #189, opened 2021-06-01
enable benchmarks

GopherJ:


yz89:
No shutdown code now. Your code means the root can add/reduce the reserve of the money market which not support yet in the current runtime.

GopherJ:
yes but in the future we will have shutdown market right @yz89

yz89 (approved):

PR #190, opened 2021-06-01
cleanup

GopherJ:
TODO:

- [x] remove 3 TODOs in primitives

companion pr: https://github.com/parallel-finance/faucet-bot/pull/6


node/parallel-dev/src/chain_spec.rs
```diff
@@ -53,10 +53,6 @@ pub fn authority_keys_from_seed(s: &str) -> (AuraId, GrandpaId) {
 
 pub fn development_config() -> Result<ChainSpec, String> {
     let wasm_binary = WASM_BINARY.ok_or_else(|| "Development wasm not available".to_string())?;
-    let mut properties = Map::new();
-    // properties.insert("tokenSymbol".into(), "PARA".into());
-    properties.insert("tokenDecimals".into(), 18.into());
```
yz89, line 58: Cloud you talk more about this?

GopherJ: I feel this is not working yet.  because yesterday when I check  polkadotjs settings it's still 12

GopherJ: also heiko & parallel didn't configure this yet. 

is it working and how to verify this btw?  because the settings still say decimals 12

yz89: Because it only is set in parallel-dev. We can have more discussion about this, and I think decimals 12 is Ok.

yz89: If you modify the `tokenDecimals`, the TOKEN_DECIMAL should as same as with the decimals in the spec settings.

primitives/src/lib.rs
```diff
@@ -96,12 +96,10 @@ pub enum CurrencyId {
     Native = 5,
 }
 
-pub const TOKEN_DECIMAL: u128 = 1_000_000_000_000_000_000;
+pub const TOKEN_DECIMAL: u128 = 1_000_000_000_000;
```
kaichaosun, line 99: We should not use this anymore.
Also don't forget to change the faucet if decimals for currencies changes.

yz89 (approved):

Issue #101, opened 2021-05-10
Pay transaction fees with KSM


GopherJ:

GopherJ:
https://github.com/paritytech/polkadot/issues/1669

GopherJ:
maybe we can use: https://github.com/paritytech/polkadot/blob/master/xcm/src/v0/order.rs#L109

GopherJ:
https://github.com/paritytech/polkadot/issues/2564#issuecomment-839724678

GopherJ:
https://github.com/paritytech/statemint/pull/51

GopherJ:
if we check here we can see it requires also fee oracle:

<img width="556" alt="image" src="https://user-images.githubusercontent.com/33961674/118396808-ce1e5f00-b683-11eb-8f9b-0ef53c00e344.png">

GopherJ:
https://github.com/parallel-finance/parallel/blob/master/runtime/heiko/src/lib.rs#L404-L413

PR #197, opened 2021-06-07
Feat/ksm native token

GopherJ:
close: #101 

- [x] update frontend code (https://github.com/parallel-finance/parallel-dapp/pull/199)

- [x] changelog

TODO in next PR:

- [ ] add runtime common & DealWithFee to make it exactly like relaychain (this requires collator_selection)


node/parallel-dev/src/chain_spec.rs
```diff
@@ -188,7 +191,7 @@ fn testnet_genesis(
                 .iter()
                 .flat_map(|x| {
                     vec![
-                        (x.clone(), CurrencyId::KSM, 10_u128.pow(21)),
+                        // (x.clone(), CurrencyId::KSM, 10_u128.pow(21)),
```
kaichaosun, line 194: Just remove it.

primitives/src/lib.rs
```diff
@@ -116,3 +107,15 @@ pub struct PriceWithDecimal {
     pub price: Price,
     pub decimal: u8,
 }
+
+////////////////////////////////////////////////////////////////////////////////
```
kaichaosun, line 111: This comment format looks weird.. please check with rust codebase.

runtime/vanilla/src/lib.rs
```diff
@@ -220,14 +225,16 @@ impl pallet_balances::Config for Runtime {
 }
 
 parameter_types! {
-    pub const TransactionByteFee: Balance = 1;
+    // 1/10 of Kusama's transaction fee
+    pub const TransactionByteFee: Balance = 1 * MILLICENTS;
 }
 
 impl pallet_transaction_payment::Config for Runtime {
+    // TODO add missing DealWithFees
```
kaichaosun, line 233: Is this still needed?

kaichaosun (approved):
A few concerns, overall looks good.

primitives/src/lib.rs
GopherJ: it's more a divider to seperate types & traits

runtime/vanilla/src/lib.rs
GopherJ: I'd like to add it, because statemint will give fees to collators so that they have some interests to run collators, but now we don't have

GopherJ: so we need to finish collator selection first and then add `DealWithFees` so that collator can receive part of fees as incentives

GopherJ:
I have no idea why coverage fails with:

```
OraclePriceNotReady
```

I need to fix it

yz89 (approved):

PR #198, opened 2021-06-07
remove block per year

GopherJ:
close: #196

### DIFF
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -110,9 +110,6 @@ pub mod module {
         /// Weight information for extrinsics in this pallet.
         type WeightInfo: WeightInfo;
 
-        /// Block per year
-        type BlockPerYear: Get<u128>;
-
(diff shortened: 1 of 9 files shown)


pallets/loans/src/lib.rs
```diff
@@ -110,9 +110,6 @@ pub mod module {
         /// Weight information for extrinsics in this pallet.
         type WeightInfo: WeightInfo;
 
-        /// Block per year
-        type BlockPerYear: Get<u128>;
-
         /// Unix time
```
ayushmishra2005, line 113: @GopherJ Can we also remove https://github.com/parallel-finance/parallel/blob/master/pallets/loans/src/rate.rs#L44

GopherJ: I'll address it after lunch

ayushmishra2005 (approved):

kaichaosun (approved):

PR #200, opened 2021-06-07
Impl Ord and PartialOrd for PriceWithDecimal

RainFallsSilent:


primitives/src/lib.rs
```diff
@@ -101,12 +101,34 @@ pub enum DataProviderId {
...
-#[derive(Encode, Decode, Debug, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Encode, Decode, Debug, Default, Copy, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 pub struct PriceWithDecimal {
     pub price: Price,
     pub decimal: u8,
 }
+impl Ord for PriceWithDecimal {
+    fn cmp(&self, other: &Self) -> Ordering {
+        if let Some((d, e)) = 10u128
```
GopherJ, line 112: maybe we can give better names than `d` and `e`

primitives/src/lib.rs
```diff
@@ -101,12 +101,34 @@ pub enum DataProviderId {
...
+            .checked_pow(self.decimal.into())
+            .zip(10u128.checked_pow(other.decimal.into()))
+        {
+            if let Some((p1, p2)) = self
```
GopherJ, line 116: maybe we can give better names than `p1` `p2`

primitives/src/lib.rs
RainFallsSilent: fixed

GopherJ (approved):

PR #204, opened 2021-06-09
[`loans`] Add `set_liquidation_incentive` call

c410-f3r:


pallets/loans/src/lib.rs
```diff
@@ -629,6 +629,31 @@ pub mod pallet {
...
+        ///
+        /// Does nothing if the provided asset is not attached to an existent incentive.
+        ///
+        /// - `currency_id`: the asset that is going to be modified.
+        #[pallet::weight(T::WeightInfo::repay_borrow_all())]
+        #[transactional]
+        pub fn set_liquidation_incentive(
+            origin: OriginFor<T>,
+            currency_id: CurrencyId,
+            liquidate_incentive: Ratio,
```
c410-f3r, line 642: Ratio is `type Ratio = Permill` and `Permill` is a wrapper around `u32`. That is why `Ratio` wasn't changed to `Rate` since both are primitives.

pallets/loans/src/lib.rs
```diff
@@ -629,6 +629,31 @@ pub mod pallet {
             Ok(().into())
         }
 
+        /// Sets a new liquidation incentive percentage for `currency_id`.
+        ///
+        /// Does nothing if the provided asset is not attached to an existent incentive.
```
c410-f3r, line 634: Not sure if this behaviour is desirable

pallets/loans/src/lib.rs
```diff
@@ -629,6 +629,31 @@ pub mod pallet {
...
+        ///
+        /// - `currency_id`: the asset that is going to be modified.
+        #[pallet::weight(T::WeightInfo::repay_borrow_all())]
```
ayushmishra2005, line 637: Shouldn't it be `set_liquidation_incentive`? Why we are applying weight for `repay_borrow_all`?

pallets/loans/src/lib.rs
```diff
@@ -629,6 +629,31 @@ pub mod pallet {
...
+            Self::ensure_currency(&currency_id)?;
+
+            let does_not_have_incentive = <LiquidationIncentive<T>>::try_get(currency_id).is_err();
+            if does_not_have_incentive {
+                return Ok(().into());
+            }
+
+            <LiquidationIncentive<T>>::insert(currency_id, liquidate_incentive);
+
+            Ok(().into())
```
ayushmishra2005, line 654: Can we remove `.into()`

pallets/loans/src/lib.rs
```diff
@@ -629,6 +629,31 @@ pub mod pallet {
...
+            origin: OriginFor<T>,
+            currency_id: CurrencyId,
+            liquidate_incentive: Ratio,
+        ) -> DispatchResultWithPostInfo {
+            ensure_root(origin)?;
+            Self::ensure_currency(&currency_id)?;
+
+            let does_not_have_incentive = <LiquidationIncentive<T>>::try_get(currency_id).is_err();
+            if does_not_have_incentive {
+                return Ok(().into());
```
ayushmishra2005, line 649: Can we remove `.into()`

node/parallel-dev/src/chain_spec.rs
```diff
@@ -28,6 +28,8 @@ use vanilla_runtime::{
     WASM_BINARY,
 };
 
+pub const DEFAULT_LIQUIDATION_INCENTIVE_PCT: Ratio = Ratio::from_percent(110);
```
kaichaosun, line 31: Here we use Ratio to represent the fixed point number range from 0 to 1, and Rate to represent the fixed point number from 0 to a big number, better use Rate if change the percent from 90 to 110, also need to update the logic that uses this number in loans pallet.

pallets/loans/src/lib.rs
```diff
@@ -629,6 +629,31 @@ pub mod pallet {
...
+        ///
+        /// - `currency_id`: the asset that is going to be modified.
+        #[pallet::weight(T::WeightInfo::repay_borrow_all())]
+        #[transactional]
+        pub fn set_liquidation_incentive(
+            origin: OriginFor<T>,
+            currency_id: CurrencyId,
+            liquidate_incentive: Ratio,
+        ) -> DispatchResultWithPostInfo {
+            ensure_root(origin)?;
```
kaichaosun, line 644: ensure_origin is more flexible.

pallets/loans/src/lib.rs
```diff
@@ -629,6 +629,31 @@ pub mod pallet {
...
+            Self::ensure_currency(&currency_id)?;
+
+            let does_not_have_incentive = <LiquidationIncentive<T>>::try_get(currency_id).is_err();
```
kaichaosun, line 647: Better to have a default value for `LiquidationIncentive` storage.

pallets/loans/src/lib.rs
yz89, line 644: we can use `UpdateOrigin` here.

pallets/loans/src/lib.rs
kaichaosun, line 649: Need to return error if no value in storage.

pallets/loans/src/tests.rs
```diff
@@ -556,6 +572,20 @@ fn ratio_and_rate_works() {
...
+fn set_liquidation_incentive_updates_stored_values() {
+    ExtBuilder::default().build().execute_with(|| {
+        let _ = Loans::set_liquidation_incentive(Origin::root(), DOT, Ratio::one());
+        let _ = Loans::set_liquidation_incentive(Origin::root(), NATIVE, Ratio::one());
+
+        assert_eq!(
+            LiquidationIncentive::<Runtime>::try_get(DOT).unwrap(),
+            Ratio::one()
+        );
+        assert!(LiquidationIncentive::<Runtime>::try_get(NATIVE).is_err());
```
kaichaosun, line 585: Why is error for native token and ok for DOT?

pallets/loans/src/tests.rs
```diff
@@ -556,6 +572,20 @@ fn ratio_and_rate_works() {
     })
 }
 
+#[test]
```
kaichaosun, line 575: Need to put related tests together.

pallets/loans/src/tests.rs
c410-f3r: Native token does not have a genesis liquidation -> https://github.com/parallel-finance/parallel/blob/ee12e6e76e32ce36ecca6c76d57993fb92e9a081/pallets/loans/src/mock.rs#L246-L251

pallets/loans/src/lib.rs
c410-f3r: `Ok(().into())` or `Ok::<_, PostDispatchInfo>`(())

pallets/loans/src/lib.rs
c410-f3r: It is written in this PR's message:
```
Still need to perform a benchmark. The call is currently using a temporary weight from T::WeightInfo::repay_borrow_all()
```

I will now run a benchmark to create the function's weight

pallets/loans/src/lib.rs
c410-f3r: A default value for the genesis configuration?

pallets/loans/src/lib.rs
c410-f3r: It is also possible to insert a new value even if the currency does not exist on `LiquidationIncentive`

node/parallel-dev/src/chain_spec.rs
c410-f3r: Thanks for the explanation. The code is now using `Rate` instead of `Ratio`

pallets/loans/src/lib.rs
ayushmishra2005: Why we can't use only `Ok(())`?

c410-f3r: `pub type DispatchResultWithPostInfo = sp_runtime::DispatchResultWithInfo<crate::weights::PostDispatchInfo>;`
`pub type DispatchResultWithInfo<T> = sp_std::result::Result<T, DispatchErrorWithPostInfo<T>>;`

`DispatchResultWithPostInfo` expects `Ok(crate::weights::PostDispatchInfo)`, thus the reason why `Ok(())` doesn't work.

`().into()` works here because of

```rust
impl From<()> for PostDispatchInfo {
	fn from(_: ()) -> Self {
		Self {
			actual_weight: None,
			pays_fee: Default::default(),
		}
	}
}
```

In any case, the call should return an actual weight instead of `PostDispatchInfo::default`.

pallets/loans/src/lib.rs
```diff
@@ -629,6 +629,30 @@ pub mod pallet {
...
+        pub fn set_liquidation_incentive(
+            origin: OriginFor<T>,
+            currency_id: CurrencyId,
+            liquidate_incentive: Rate,
+        ) -> DispatchResultWithPostInfo {
+            T::UpdateOrigin::ensure_origin(origin)?;
+            Self::ensure_currency(&currency_id)?;
+
+            if <LiquidationIncentive<T>>::try_get(currency_id).is_err() {
+                return Err(<Error<T>>::CurrencyNotEnabled.into());
```
kaichaosun, line 648: into is not needed here.

pallets/loans/src/lib.rs
```diff
@@ -629,6 +629,30 @@ pub mod pallet {
...
+            }
+
+            <LiquidationIncentive<T>>::insert(currency_id, liquidate_incentive);
+
+            Ok(Some(T::WeightInfo::set_liquidation_incentive()).into())
```
kaichaosun, line 653: There is no need to do this, based on the doc
```
/// The return typ of a `Dispatchable` in frame. When returned explicitly from
/// a dispatchable function it allows overriding the default `PostDispatchInfo`
/// returned from a dispatch.
```

pallets/loans/src/tests.rs
kaichaosun: I mean tests cases near to each other, but still seperate each case with a standalone function.

kaichaosun (commented):
There is also a bench bot which can run the benchmark with a preconfigured virtual machine. Check out the past PRs.

pallets/loans/src/lib.rs
c410-f3r: Nice! Thank you

pallets/loans/src/lib.rs
c410-f3r: To convert `pallet::Error` into `DispatchErrorWithPostInfo`, it is necessary to call `into()` like in https://github.com/parallel-finance/parallel/blob/ee12e6e76e32ce36ecca6c76d57993fb92e9a081/pallets/loans/src/lib.rs#L1215

c410-f3r:
```
thread 'tests::borrow_works' panicked at 'Expected Ok(_). Got Err(
    DispatchErrorWithPostInfo {
        post_info: PostDispatchInfo {
            actual_weight: None,
            pays_fee: Pays::Yes,
        },
        error: Module {
            index: 4,
            error: 3,
            message: Some(
                "PriceOracleNotReady",
            ),
        },
    },
)', pallets/loans/src/tests.rs:145:9
```

Error doesn't seem related. Let me double check it

GopherJ:
> ```
> thread 'tests::borrow_works' panicked at 'Expected Ok(_). Got Err(
>     DispatchErrorWithPostInfo {
>         post_info: PostDispatchInfo {
>             actual_weight: None,
>             pays_fee: Pays::Yes,
>         },
>         error: Module {
>             index: 4,
>             error: 3,
>             message: Some(
>                 "PriceOracleNotReady",
>             ),
>         },
>     },
> )', pallets/loans/src/tests.rs:145:9
> ```
> 
> Error doesn't seem related. Let me double check it

it's CI issue, I clicked re-run for you. I'll fix it once got time

ayushmishra2005 (approved):

node/parallel-dev/src/chain_spec.rs
```diff
@@ -214,9 +217,9 @@ fn testnet_genesis(
                 (CurrencyId::xKSM, Ratio::from_percent(90)),
             ],
             liquidation_incentive: vec![
-                (CurrencyId::KSM, Ratio::from_percent(90)),
-                (CurrencyId::USDT, Ratio::from_percent(90)),
-                (CurrencyId::xKSM, Ratio::from_percent(90)),
+                (CurrencyId::KSM, DEFAULT_LIQUIDATION_INCENTIVE_PCT),
```
kaichaosun, line 220: You have changed the value from 90% to 110%, but haven't change the place that uses this storage, the logic should be different when using the new value, can you check this?

yz89: before: collateral_value * incentive(90%) = repay_value
expect: collateral_value = repay_value * incentive(110%)

c410-f3r:
After researching about the subject, it is possible that now I finally understood what needs to be done to accomplish the final goal. Please, feel free to inform if something is still missing.

The number of commits were getting too big so I squashed them.

node/parallel-dev/src/chain_spec.rs
```diff
@@ -28,6 +28,9 @@ use vanilla_runtime::{
     WASM_BINARY,
 };
 
+// 1.100000000000000000 or 110%
+pub const DEFAULT_LIQUIDATION_INCENTIVE_PCT: Rate = Rate::from_inner(Rate::DIV / 100 * 110);
```
yz89, line 32: I think `Rate::saturating_from_rational(110, 100)` is better.

yz89:
Nice work! You have done what need to do. 👍

c410-f3r:
Thank you very much for your help @yz89!

It is pretty late now so I will convert `pub const DEFAULT_LIQUIDATION_INCENTIVE_PCT: Rate = Rate::from_inner(Rate::DIV / 100 * 110);` into locals `Rate::saturating_from_rational(110, 100)` tomorrow morning.

yz89:
Sure, good night.

PR #213, opened 2021-06-14
Change origins for loan pallet according to Governance Module

ayushmishra2005:


GopherJ (approved):

runtime/heiko/src/lib.rs
```diff
@@ -265,8 +265,8 @@ impl pallet_loans::Config for Runtime {
     type Currency = Currencies;
     type PalletId = LoansPalletId;
     type PriceFeeder = Prices;
-    type ReserveOrigin = EnsureRoot<AccountId>;
-    type UpdateOrigin = EnsureRoot<AccountId>;
+    type ReserveOrigin = EnsureRootOrHalfCouncil;
```
GopherJ, line 268: I think 2/3 is better

kaichaosun (approved):

runtime/heiko/src/lib.rs
ayushmishra2005: @kaichaosun @yz89  What do you suggest here?

yz89: We can just use `EnsureRootOrHalfCouncil` now because it has been used by several modules.

Issue #212, opened 2021-06-14
integrate xtokens back


GopherJ:

GopherJ:
https://github.com/open-web3-stack/open-runtime-module-library/tree/sw/the-return-of-the-xcm

PR #215, opened 2021-06-14
bring xtokens back

GopherJ:
close: #212 

- [x] test if downward/upward transfer still works with KSM/DOT
- [ ] add upward transfer for liquid-staking withdraw function


GopherJ:
close in favor of native one

PR #223, opened 2021-06-17
Add more units tests to attest edge cases involving errors - Part 1

c410-f3r:
Changes are becoming quite large so this PR only brings a subset of all required additional unit tests for an easier review. More tests will be added afterwards.

Code should be read commit-by-commit for a better understanding of what is going on.

```
    Add more units tests to attest edge cases involving errors
    
    Beyond the addition of more tests, also moves very large common errors into unified modules.
```

```
    Use built-in overflow errors
    
    Substrate already provides built-in errors for arithmetic errors so custom error enum variants are unnecessary
```

```
    Re-arrange modules, exports and cfg flags
    
    `cfg` flags for modules can be placed inside the module itself enhacing overall readability.
    Also groups scattered exports and modules declarations.
 ```

### DIFF
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -34,9 +34,11 @@ use frame_support::{
 };
 use frame_system::pallet_prelude::*;
 use orml_traits::{MultiCurrency, MultiCurrencyExtended};
+pub use pallet::*;
 use primitives::{
     Amount, Balance, CurrencyId, Multiplier, Price, PriceFeeder, Rate, Ratio, Timestamp,
 };
+use sp_runtime::ArithmeticError;
 use sp_runtime::{
     traits::{
         AccountIdConversion, CheckedAdd, CheckedDiv, CheckedMul, CheckedSub, StaticLookup, Zero,
@@ -45,18 +47,12 @@ use sp_runtime::{
 };
 use sp_std::result;
 use sp_std::vec::Vec;
-
-pub use pallet::*;
-
-pub mod weights;
-
 pub use weights::WeightInfo;
 
-#[cfg(test)]
 mod mock;
 mod rate;
-#[cfg(test)]
 mod tests;
+pub mod weights;
 
 /// Container for borrow balance information
 #[derive(Encode, Decode, Eq, PartialEq, Copy, Clone, RuntimeDebug, Default)]
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -127,7 +127,7 @@ pub mod pallet {
         /// Insufficient collateral asset to borrow more or disable collateral
         InsufficientCollateral,
         /// Repay amount greater than borrow balance
-        RepayAmountTooBig,
+        RepayAmountExceedsCloseFactor,
         /// Asset already enabled/disabled collateral
         DuplicateOperation,
         /// No deposit asset
@@ -952,7 +952,7 @@ impl<T: Config> Pallet<T> {
     ) -> DispatchResult {
         let account_borrows = Self::borrow_balance_stored(borrower, currency_id)?;
         if account_borrows < repay_amount {
-            return Err(Error::<T>::RepayAmountTooBig.into());
+            return Err(Error::<T>::RepayAmountExceedsCloseFactor.into());
         }
 
         T::Currency::transfer(*currency_id, borrower, &Self::account_id(), repay_amount)?;
@@ -1055,7 +1055,7 @@ impl<T: Config> Pallet<T> {
         // we can only liquidate 50% of the borrows
         let close_factor = CloseFactor::<T>::get(liquidate_currency_id);
         if close_factor.mul_floor(account_borrows) < repay_amount {
-            return Err(Error::<T>::RepayAmountTooBig.into());
+            return Err(Error::<T>::RepayAmountExceedsCloseFactor.into());
         }
 
         //calculate collateral_token_sum price
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -54,6 +54,9 @@ mod rate;
 mod tests;
 pub mod weights;
 
+pub(crate) const OVERFLOW: DispatchError = DispatchError::Arithmetic(ArithmeticError::Overflow);
+pub(crate) const UNDERFLOW: DispatchError = DispatchError::Arithmetic(ArithmeticError::Overflow);
+
 /// Container for borrow balance information
 #[derive(Encode, Decode, Eq, PartialEq, Copy, Clone, RuntimeDebug, Default)]
 pub struct BorrowSnapshot {
@@ -143,10 +146,6 @@ pub mod pallet {
         InvalidRateModelParam,
         /// Currency not enabled
         CurrencyNotEnabled,
-        /// Calculation overflow
-        Overflow,
-        /// Calculation underflow
-        Underflow,
         /// Currency's oracle price not ready
         PriceOracleNotReady,
     }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by c410-f3r** [comment]
At least locally, tests are very flaky and I am not sure what is causing this intermittent behaviour.

File: pallets/loans/src/lib.rs [hunk 1]
```diff
@@ -1332,11 +1306,11 @@ impl<T: Config> Pallet<T> {
     pub fn calc_collateral_amount(
         underlying_amount: u128,
         exchange_rate: Rate,
-    ) -> result::Result<Balance, Error<T>> {
+    ) -> result::Result<Balance, DispatchError> {
```
GopherJ, line 1309: result:: is not necessary here

pallets/loans/src/lib.rs
```diff
@@ -45,18 +47,15 @@ use sp_runtime::{
...
 
-#[cfg(test)]
 mod mock;
 mod rate;
-#[cfg(test)]
 mod tests;
+pub mod weights;
+
+pub(crate) const OVERFLOW: DispatchError = DispatchError::Arithmetic(ArithmeticError::Overflow);
+pub(crate) const UNDERFLOW: DispatchError = DispatchError::Arithmetic(ArithmeticError::Overflow);
```
GopherJ, line 58: we can remove `pub(crate)` if this will only be used by this pallet, if we would like to keep it we can move it to a common place

also this should be
```
 pub(crate) const UNDERFLOW: DispatchError = DispatchError::Arithmetic(ArithmeticError::Underflow);
 ```
right?

File: pallets/loans/src/lib.rs [hunk 3]
```diff
@@ -1010,7 +990,7 @@ impl<T: Config> Pallet<T> {
     pub fn borrow_balance_stored_with_snapshot(
         currency_id: &CurrencyId,
         snapshot: BorrowSnapshot,
-    ) -> result::Result<Balance, Error<T>> {
+    ) -> result::Result<Balance, DispatchError> {
```
GopherJ, line 993: I forgot why do we need to type `result::Result` instead of `Result`

pallets/loans/src/lib.rs
yz89: `ArithmeticError` already implemented the `From` for `DispatchError`. I don't think it's necessary to type the const one.
https://github.com/paritytech/substrate/blob/master/primitives/runtime/src/lib.rs#L593

yz89: We can use `ArithmeticError` directly like this. https://github.com/AcalaNetwork/Acala/blob/master/modules/loans/src/lib.rs#L325

pallets/loans/src/lib.rs
yz89: Because we can not use `std::Result` in wasm runtime. 

`sp_std::result` is provided by substrate.

GopherJ: then we can import sp_std::result::Result maybe

yz89: That makes sense.

pallets/loans/src/tests/liquidate_borrow.rs
```diff
@@ -0,0 +1,128 @@
...
+    ExtBuilder::default().build().execute_with(|| {
+        initial_setup();
+        alice_borrows_100_ksm();
+        MOCK_PRICE_FEEDER::set_price(KSM, Rate::from_float(2000.0));
+        LiquidationIncentive::<Runtime>::insert(KSM, Rate::from_float(200.0));
+        assert_noop!(
+            Loans::liquidate_borrow(Origin::signed(BOB), ALICE, KSM, million_dollar(50), DOT),
+            Error::<Runtime>::RepayValueGreaterThanCollateral
+        );
+    })
```
yz89, line 33: MOCK_PRICE_FEEDER::reset();

pallets/loans/src/tests/liquidate_borrow.rs
```diff
@@ -0,0 +1,128 @@
...
+    ExtBuilder::default().build().execute_with(|| {
+        initial_setup();
+        alice_borrows_100_ksm();
+        MOCK_PRICE_FEEDER::set_price(KSM, 20.into());
+        assert_noop!(
+            Loans::liquidate_borrow(Origin::signed(BOB), ALICE, KSM, million_dollar(51), DOT),
+            Error::<Runtime>::RepayAmountExceedsCloseFactor
+        );
+    })
+}
```
yz89, line 101: MOCK_PRICE_FEEDER::reset();

pallets/loans/src/lib.rs
```diff
@@ -34,9 +34,11 @@ use frame_support::{
 };
 use frame_system::pallet_prelude::*;
 use orml_traits::{MultiCurrency, MultiCurrencyExtended};
+pub use pallet::*;
```
ayushmishra2005, line 37: Why do we need this?

kaichaosun: Runtime needs to find the data types defined in this module.

pallets/loans/src/tests/liquidate_borrow.rs
```diff
@@ -0,0 +1,128 @@
...
+        initial_setup();
+        assert_noop!(
+            Loans::liquidate_borrow(Origin::signed(BOB), ALICE, KSM, 0, DOT),
+            Error::<Runtime>::NoBorrowBalance
+        );
+    })
+}
+
+#[test]
+fn collateral_value_must_be_less_than_liquidation_value() {
```
kaichaosun, line 23: less -> bigger

c410-f3r: Done

pallets/loans/src/tests/liquidate_borrow.rs
c410-f3r: Included

pallets/loans/src/lib.rs
c410-f3r: Modified

pallets/loans/src/lib.rs
c410-f3r: Removed in favor of local declarations

yz89:
> At least locally, tests are very flaky and I am not sure what is causing this intermittent behavior.
> 
> EDIT: Probably because of modified storage parameters. I will investigate tomorrow.

It mostly caused by `MOCK_PRICE_FEEDER` data race. Let's consider how to use `MOCK_PRICE_FEEDER` in a multi-thread environment.

pallets/loans/src/tests.rs
```diff
@@ -934,3 +898,14 @@ fn get_price_works() {
...
+// Groups all tests that conflict internal storage due to concurrent reads and writes
+// or in other words, groups all flaky concurrent tests.
+#[test]
+fn sequential_tests() {
+    total_collateral_asset_value_works();
+
+    liquidate_borrow::collateral_value_must_be_greater_than_liquidation_value();
+    liquidate_borrow::full_workflow_works_as_expected();
+    liquidate_borrow::liquidator_can_not_repay_more_than_the_close_factor_pct_multiplier();
+}
```
c410-f3r, line 911: `cargo test` spawns `N`, where `N = number of host cores`, jobs so parallel by default and by design.
To avoid having to create a sequential entry-point for tests, it is also possible to always run `cargo test -- --test-threads 1`

yz89: That makes sense. Maybe this will fix the intermittent problems in Test Coverage. @GopherJ

yz89 (approved):

Issue #222, opened 2021-06-17
add parachain binary to docker image


GopherJ:

GopherJ:
now our docker image: `parallelfinance/parallel:latest` contains only vanilla node, we should test it well and put some docs in README.md

then we also need to add parachain binary to make it possible to run a collator using our docker image

GopherJ:
acala has their own launch module which helps launching relaychain & parachain in production mode

https://github.com/AcalaNetwork/Acala/tree/master/launch


polkadot-launch can only be used for testing purpose but it's not good for production usage as it just spawns processes

Issue #195, opened 2021-06-02
liquidation message is confusing


GopherJ:
<img width="1055" alt="image" src="https://user-images.githubusercontent.com/33961674/120505880-fdf59280-c3f7-11eb-8433-47709103a79c.png">

it outputs `liquidation successful` without real operation

Issue #185, opened 2021-06-01
add scripts/github-action to use srtool


GopherJ:

GopherJ:
chevdor/srtool-actions

Issue #181, opened 2021-05-31
opt-in/opt-out parallel & heiko to reduce wasm size


GopherJ:
currently it took pretty much time for parachain to produce blocks like 30seconds, which is not normal, I'm not sure if it's caused by two runtime (heiko-runtime, parallel-runtime)

GopherJ:
if we can solve the block producing timing issue, then it's not necessay... also shell takes also 12 seconds to produce block, so maybe it's just caused by polkadot

PR #225, opened 2021-06-19
integrate srtool

GopherJ:
close: #222 
close: #185 
close: #195 

download wasm here: https://github.com/parallel-finance/parallel/actions/runs/952074365

- [ ] test if the wasm works with collator built locally


GopherJ:
image

the two runtime have the same `hash` which isn't normal, I think maybe we need to finish #181 to seperate these two runtime instead of building one with both `heiko` and `parallel`

ayushmishra2005 (approved):

node/parallel/src/client.rs
```diff
@@ -176,7 +180,9 @@ impl ClientHandle for Client {
 impl sc_client_api::UsageProvider<Block> for Client {
     fn usage_info(&self) -> sc_client_api::ClientInfo<Block> {
         match self {
+            #[cfg(feature = "with-parallel-runtime")]
```
kaichaosun, line 183: I remember we discussed that won't increase code complexity with these features for build speedup.

GopherJ: looks like the only difference is `WithdrawAmount`  config

GopherJ: I reverted that commit for now

PR #226, opened 2021-06-19
refactor rate model to support curve model

yz89:
- remove APR struct
- modify the `InterestRateModel` type from struct to enum
- support curve rate model


pallets/loans/src/lib.rs
```diff
@@ -356,13 +356,18 @@ pub mod pallet {
...
                     self.base_rate,
-                    self.kink_rate,
+                    self.jump_rate,
                     self.full_rate,
-                    self.kink_utilization,
+                    self.jump_utilization,
                 );
-                interest_model.check_parameters().unwrap();
+                if !interest_model.check_model() {
+                    log::info!(
```
ayushmishra2005, line 366: Shouldn't this be log.error?

node/parallel/src/chain_spec/heiko.rs
```diff
@@ -185,9 +185,9 @@ fn testnet_genesis(
             borrow_index: Rate::one(),                             // 1
             exchange_rate: Rate::saturating_from_rational(2, 100), // 0.02
             base_rate: Rate::saturating_from_rational(2, 100),     // 2%
-            kink_rate: Rate::saturating_from_rational(10, 100),    // 10%
+            jump_rate: Rate::saturating_from_rational(10, 100),    // 10%
```
kaichaosun, line 188: These model field can be grouped use a struct or tuple, can be reactored later.

kaichaosun (approved):
👍

c410-f3r (approved):
`pallets/loans/src/rate.rs` was transposed to the `InterestRateModel::Jump` enum variant so it should only be a matter of implement the remaining functions for `InterestRateModel::CurveModel`.

Thanks for doing this introductory work for #227

yz89:
> `pallets/loans/src/rate.rs` was transposed to the `InterestRateModel::Jump` enum variant so it should only be a matter of implement the remaining functions for `InterestRateModel::CurveModel`.
> 
> Thanks for doing this introductory work for #227

Yes, you are right!

GopherJ (approved):

node/parallel/src/chain_spec/heiko.rs
yz89: Sure

pallets/loans/src/lib.rs
yz89: Thanks, I've replaced it with panic.

Issue #229, opened 2021-06-21
Modify the Interest Rate Curve and Add Interest Bearing Token


yubo-ruan:
**Quick Issue**
We use the exact same interest rate model parameters from Compound chain. I think we should change it as soon as possible. 

**Suggested Solution**

1. Change name to 
base_rate: Rate
max_rate: Rate
optimal_rate: Rate
optimal_utilization: Ratio

**New Curve Motivation**

Currently, we are using the classic piecewise linear function from the Compound interest rate model. However, it is not optimized for KSM pool because of the KSM staking rewards. I will explain why in the next two paragraphs. 

Notice, staking rewards will compete with KSM lending pool. If the KSM staking rate is higher than the lending rate, users have the incentive to withdraw and stake. If staking rate is lower, users have the incentive to unstake and supply to KSM pool.

Since we will have leverage staking functionality, as long as the borrowing rate is lower than the staking rate, users have an incentive to borrow and leverage up.  Similarly, as long as the lending rate is higher than the staking rate, users have an incentive to lend in our lending poo. In the end, the equilibrium will be reached when the borrowing rate is higher than the staking rate (no one wants to borrow), and the lending rate is lower than the staking rate (no one want to supply). 

2. Peg the optimal_rate to the staking_apy (currently 14%)

Based on the above reasons, we should put optimal_rate in the middle of supply_rate and borrow_rate
optimal_rate = staking_apy + interest_rate_distance/2
interest_rate_distance = borrow_rate - supply_rate = borrow_rate * [1 - utilization * (1- reserveRatio)]
Therefore: 
optimal_rate = staking_apy +  borrow_rate * [1 - utilization * (1- reserveRatio)] / 2

Now the question is how to calculate the dynamic staking rate. I suggest using 14% to get started but we should peg to actual staking apy. 

staking_apy = xKSM_exchange_rate / xKSM_exchange_rate(last_week) * 52.1429 

**Alternatives**

We can use the exponential curve that is on our white paper, but it is not the highest priority and we should do more research before we use it. 

**Interest-bearing Token Motivation**

Compatibility is one of the most important factors that make DeFi scale and grows. We should make our money market be one of the financial primitives for our users. We want to add interest-bearing tokens to our users. 

Potently use case includes supplying our interest bearing token to AMM or yield aggregator. (check out https://curve.fi/aave)

**Suggested Solution**

[...]

PR #231, opened 2021-06-22
New rate model

c410-f3r:
The current state is not meant to be the final version of #229 but serve as a discussion platform for the desired complete set of changes.

If appropriated and necessary, individual commits can be extracted into different PRs.

This PR can be better read commit-by-commit.

### DIFF
--- a/pallets/loans/benchmarking/src/lib.rs
+++ b/pallets/loans/benchmarking/src/lib.rs
@@ -10,7 +10,7 @@ use frame_system::RawOrigin as SystemOrigin;
 use orml_oracle::Instance1;
 use orml_oracle::{Config as ORMLOracleConfig, Pallet as ORMLOracle};
 use orml_traits::MultiCurrency;
-use pallet_loans::{Config as LoansConfig, InterestRateModel, Pallet as Loans};
+use pallet_loans::{Config as LoansConfig, InterestRateModels, Pallet as Loans};
 use primitives::{CurrencyId, PriceWithDecimal, Rate, Ratio};
 use sp_runtime::traits::{Bounded, One, StaticLookup};
 use sp_runtime::{ArithmeticError, FixedPointNumber, FixedU128};
@@ -289,7 +289,7 @@ benchmarks! {
          let _ = Loans::<T>::set_rate_model(
             SystemOrigin::Root.into(),
             DOT,
-            InterestRateModel::new_jump_model(
+            InterestRateModels::new_jump_model(
                 Rate::saturating_from_rational(5, 100),
                 Rate::saturating_from_rational(15, 100),
                 Rate::saturating_from_rational(35, 100),
@@ -299,7 +299,7 @@ benchmarks! {
     verify {
         assert_eq!(
             Loans::<T>::currency_interest_model(DOT),
-            InterestRateModel::new_jump_model(
+            InterestRateModels::new_jump_model(
                 Rate::saturating_from_rational(5, 100),
                 Rate::saturating_from_rational(15, 100),
                 Rate::saturating_from_rational(35, 100),
--- a/pallets/loans/benchmarking/src/lib.rs
+++ b/pallets/loans/benchmarking/src/lib.rs
@@ -10,7 +10,7 @@ use frame_system::RawOrigin as SystemOrigin;
 use orml_oracle::Instance1;
 use orml_oracle::{Config as ORMLOracleConfig, Pallet as ORMLOracle};
 use orml_traits::MultiCurrency;
-use pallet_loans::{Config as LoansConfig, InterestRateModels, Pallet as Loans};
+use pallet_loans::{Config as LoansConfig, InterestRateModel, Pallet as Loans};
 use primitives::{CurrencyId, PriceWithDecimal, Rate, Ratio};
 use sp_runtime::traits::{Bounded, One, StaticLookup};
 use sp_runtime::{ArithmeticError, FixedPointNumber, FixedU128};
@@ -289,7 +289,7 @@ benchmarks! {
          let _ = Loans::<T>::set_rate_model(
             SystemOrigin::Root.into(),
             DOT,
-            InterestRateModels::new_jump_model(
+            InterestRateModel::new_jump_model(
                 Rate::saturating_from_rational(5, 100),
                 Rate::saturating_from_rational(15, 100),
                 Rate::saturating_from_rational(35, 100),
@@ -299,7 +299,7 @@ benchmarks! {
     verify {
         assert_eq!(
             Loans::<T>::currency_interest_model(DOT),
-            InterestRateModels::new_jump_model(
+            InterestRateModel::new_jump_model(
                 Rate::saturating_from_rational(5, 100),
                 Rate::saturating_from_rational(15, 100),
                 Rate::saturating_from_rational(35, 100),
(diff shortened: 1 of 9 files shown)


pallets/loans/src/rate_model.rs
```diff
@@ -13,18 +13,27 @@
...
+}
+
 /// Parallel interest rate model
 #[derive(Encode, Decode, Eq, PartialEq, Copy, Clone, RuntimeDebug)]
-pub enum InterestRateModel {
-    Jump(JumpModel),
+pub enum InterestRateModels {
     Curve(CurveModel),
+    Jump(JumpModel),
+    NewCurveName(NewCurveNameModel),
```
c410-f3r, line 33: It is necessary to replace `NewCurveName` with the actual name of the new model (Couldn't figure out one)

pallets/loans/src/rate_model.rs
```diff
@@ -165,19 +185,57 @@ impl CurveModel {
...
+            borrow_rate,
+            reserve_ratio,
+            staking_apy,
+        }
+    }
+}
+
+impl InterestRateModel for NewCurveNameModel {
+    fn check_model(&self) -> bool {
+        false
```
c410-f3r, line 219: Also couldn't figure out sane values for the struct fields. Any suggestion is welcome

pallets/loans/benchmarking/src/lib.rs
```diff
@@ -289,7 +289,7 @@ benchmarks! {
          let _ = Loans::<T>::set_rate_model(
             SystemOrigin::Root.into(),
             DOT,
-            InterestRateModel::new_jump_model(
+            InterestRateModels::new_jump_model(
```
GopherJ, line 292: I think we don't need to rename

pallets/loans/src/rate_model.rs
```diff
@@ -13,18 +13,27 @@
 // limitations under the License.
 
 use primitives::{Rate, Ratio, Timestamp, SECONDS_PER_YEAR};
-use sp_runtime::traits::{CheckedAdd, CheckedDiv, CheckedSub, Saturating};
+use sp_runtime::traits::{CheckedAdd, CheckedDiv, CheckedMul, CheckedSub, Saturating};
 
 use crate::*;
 
+pub trait InterestRateModel {
```
GopherJ, line 20: you can name it:

```
InterestRateModelLike
```

pallets/loans/src/rate_model.rs
```diff
@@ -165,19 +185,57 @@ impl CurveModel {
...
+    reserve_ratio: Ratio,
+    staking_apy: Ratio,
+}
+
+impl NewCurveNameModel {
+    pub fn new(borrow_rate: Rate, reserve_ratio: Ratio, staking_apy: Ratio) -> Self {
+        Self {
+            borrow_rate,
+            reserve_ratio,
+            staking_apy,
```
c410-f3r, line 212: As described on #229, `staking_apy` should be `xKSM_exchange_rate / xKSM_exchange_rate(last_week) * 52.1429` but it is currently a percentage because the semantic of `52.1429` is unknown. Or maybe it is just me that couldn't spot its meaning.
Please, feel free to discuss which direction should `staking_apy` take.

pallets/loans/src/rate_model.rs
```diff
@@ -259,11 +320,26 @@ mod tests {
...
+    #[test]
+    fn new_curve_name_correctly_calculates_borrow_rate() {
+        let new_curve_name_model = NewCurveNameModel::new(
+            Rate::from_inner(Rate::DIV / 100 * 110), // 110%
+            Ratio::from_percent(50),
+            Ratio::from_percent(14),
+        );
+        assert_eq!(
+            new_curve_name_model
+                .get_borrow_rate(Ratio::from_percent(10))
```
c410-f3r, line 340: These method parameters were gathered out of the blue and don't have actual or reasonable values. They were used to simply verify that the calculation is correct.

ayushmishra2005 (approved):

pallets/loans/src/rate_model.rs
```diff
@@ -79,75 +81,93 @@ impl InterestRateModel {
...
+    pub const MAX_OPTIMAL_RATE: Rate = Rate::from_inner(500_000_000_000_000_000); // 50%
 
     /// Create a new rate model
     pub fn new_model(
         base_rate: Rate,
-        jump_rate: Rate,
-        full_rate: Rate,
-        jump_utilization: Ratio,
+        max_rate: Rate,
+        optimal_rate: Rate,
```
yz89, line 124: jump_rate should be renamed to optimal_rate.
full_rate should be renamed to max_rate.

yz89: MAX_RATE and MAX_OPTIMAL_RATE need a corresponding modification.

c410-f3r: Done

pallets/loans/src/rate_model.rs
yz89: staking_apy should come from liquid_staking pallet, and it has not been implemented yet. We can just use the const 14% for now.

pallets/loans/src/rate_model.rs
```diff
@@ -165,19 +185,57 @@ impl CurveModel {
...
+    fn get_borrow_rate(&self, _: Ratio) -> Option<Rate> {
         None
     }
 }
 
+#[derive(Clone, Copy, Decode, Default, Encode, Eq, PartialEq, RuntimeDebug)]
+pub struct NewCurveNameModel {
+    borrow_rate: Rate,
+    reserve_ratio: Ratio,
+    staking_apy: Ratio,
```
yz89, line 204: reserve_ratio should use the ReserveFactor storage.

pallets/loans/src/rate_model.rs
yubo-ruan: I think we can call it **PolynomialCurve** since the new curve has more than two algebraic terms, and it is the sum of several terms that contain different powers of the same variable. @c410-f3r

pallets/loans/src/rate_model.rs
yubo-ruan: See above comment

pallets/loans/src/rate_model.rs
yubo-ruan: I agree. We can use a const for now but highly suggest that we can have an interface to export the current staking_apy

pallets/loans/src/rate_model.rs
c410-f3r: Thank you

pallets/loans/src/rate_model.rs
c410-f3r: Renamed to `reserve_ratio` to remember that its value should come from the `ReserveFactor` storage

pallets/loans/src/rate_model.rs
c410-f3r: Changed `staking_apy` to a constant

c410-f3r:
@yz89 As requested, the new curve model is now using the jump model algorithm with `staking_apy` as the `optimal_utilization` parameter

pallets/loans/src/rate_model.rs
kaichaosun: The interest rate is dynamic, it's related to the staking apy at product level, not the code/protocol level.
Make it configurable from runtime makes more sense.

c410-f3r: IIUC, `interest_rate == borrow_rate` parameter retrieved from a runtime configuration, therefore, a new `Get<Rate>` type was added to `loans::Config`.

yz89:
Close this PR temporarily, for we need more discussion about #229.

PR #235, opened 2021-06-22
change liquid staking's initial exchange rate to be 1

GopherJ:


.github/workflows/deploy.yml
```diff
@@ -38,12 +38,11 @@ jobs:
 
       - name: Build Release And Package
         run: |
-          cargo build --release
+          cargo build --bin parallel-dev --release
```
GopherJ, line 41: I changed this back because https://github.com/parallel-finance/parallel/actions/runs/959539598 means that github actions have difficuties to build cumulus

yubo-ruan:
Looks good to me

PR #239, opened 2021-06-23
add redeem_allowed to protect excess withdraw

yz89:


GopherJ:
I think:

```
ensure_borrow_allowed
ensure_redeem_allowed
```

will be better, not sure how the others think

GopherJ (approved):

pallets/loans/src/lib.rs
```diff
@@ -833,37 +835,22 @@ impl<T: Config> Pallet<T> {
...
-
-        Ok(())
+    fn get_account_liquidity(
+        account: &T::AccountId,
+    ) -> Result<(FixedU128, FixedU128), DispatchError> {
+        let total_borrow_value = Self::total_borrowed_value(account)?;
+        let total_collateral_value = Self::total_collateral_value(account)?;
+        if total_collateral_value > total_borrow_value {
+            Ok((
+                total_collateral_value - total_borrow_value,
```
c410-f3r, line 845: This line probably won't overflow in practice but maybe play safe and use `total_collateral_value.checked_sub(total_borrow_value).or_ok(ArithmeticError::Underflow)?` instead?

c410-f3r (approved):

pallets/loans/src/lib.rs
```diff
@@ -940,17 +909,85 @@ impl<T: Config> Pallet<T> {
...
+            return Err(Error::<T>::InsufficientDeposit.into());
+        }
+        if !deposit.is_collateral {
+            return Ok(());
+        }
+        let collateral_factor = Self::collateral_factor(currency_id);
+        let price = Self::get_price(currency_id)?;
+        let redeem_effects_value = price
+            .checked_mul(&FixedU128::from_inner(
+                collateral_factor.mul_ceil(redeem_amount),
```
GopherJ, line 931: why do we use `mul_ceil` here?

GopherJ, line 931: 👍

pallets/loans/src/lib.rs
yz89: I don't think it's necessary for It's an easy case.

PR #240, opened 2021-06-23
fix some issues of polkadot-v0.9.5

GopherJ:
polkadot-launch v1.6.3 should work if https://github.com/paritytech/polkadot-launch/pull/117 gets merged and released

otherwise we can also use `make register-parachain` to register our parachain

### DIFF
--- a/runtime/heiko/src/constants.rs
+++ b/runtime/heiko/src/constants.rs
@@ -16,7 +16,7 @@ pub mod currency {
     use primitives::Balance;
     pub const EXISTENTIAL_DEPOSIT: u128 = 500;
 
-    pub const MILLICENTS: Balance = 1_000_000_000;
+    pub const MILLICENTS: Balance = 10_000_000;
     pub const CENTS: Balance = 1_000 * MILLICENTS; // assume this is worth about a cent.
     pub const DOLLARS: Balance = 100 * CENTS;
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5344,7 +5344,6 @@ version = "3.0.0"
 dependencies = [
  "frame-support",
  "frame-system",
- "lazy_static",
  "orml-currencies",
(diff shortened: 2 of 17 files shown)


runtime/heiko/src/constants.rs
```diff
@@ -16,7 +16,7 @@ pub mod currency {
     use primitives::Balance;
     pub const EXISTENTIAL_DEPOSIT: u128 = 500;
 
-    pub const MILLICENTS: Balance = 1_000_000_000;
+    pub const MILLICENTS: Balance = 10_000_000;
```
c410-f3r, line 19: I am assuming that these currency parameters and balance storage are related. Feel free to ignore this comment if they are not.

If a millicent is 10^-3 cent and a cent is 10^-3 of an integral unit of value, then a integral unit of value is 10^-6 millicents or `0.000001 * 1000000`.
Parallel uses `FixedU128` that has 18 decimals so a milicent should be 10^12 instead of 10^7

```rust
let _ = FixedU128::from_inner(FixedU128::DIV / 1_000_000); // 0.000001
```

IIRC, `10_000_000` is related to the number of decimals used by one of the parity's runtimes

c410-f3r (approved):

runtime/heiko/src/constants.rs
GopherJ: I think `FixedU128` is not related to decimals, substrate's default decimal is 12, so I think I just did it like this, the previous one was a bit strange for me

GopherJ:
@c410-f3r @yz89  can you help checking if https://github.com/parallel-finance/parallel/pull/240/commits/5d7f3d52470707e15c7908ea172055ec29128a7d will fix test coverage issue? I'm not sure yet but would like to try and sorry that I didn't open another pr for this:)

GopherJ:
let's try first:)

PR #245, opened 2021-06-25
Staking Pallet benchmarking

ayushmishra2005:


pallets/liquid-staking/benchmarking/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+        assert_eq!(processing_unstake[0].amount, amount);
+        assert_eq!(
+            processing_unstake[0].block_number,
+            frame_system::Pallet::<T>::block_number()
+        );
+
+        assert_last_event::<T>(pallet_liquid_staking::Event::UnstakeProcessing(agent, caller, amount).into());
+    }
+
+    finish_processed_unstake {
```
kaichaosun, line 205: There ma ybe a lot of items in collections, which means we need to benchmark the worst case. Here you just benchmark one stake and one unstake. If the worst case is not able to find, we need to use bounded collection.
Please check this call and above calls.

GopherJ (approved):

pallets/liquid-staking/src/benchmarking.rs
```diff
@@ -0,0 +1,251 @@
...
+        initial_set_up::<T>(caller.clone());
+        let agent: T::AccountId = account("Sample", 6, SEED);
+        let amount = 100_000;
+        let withdraw_amount = T::MaxWithdrawAmount::get() - 10;
+        assert_ok!(LiquidStaking::<T>::stake(
+            SystemOrigin::Signed(caller.clone()).into(),
+            amount));
+        let call = Call::<T>::withdraw(agent.clone(), withdraw_amount);
+        let origin = SystemOrigin::Root.into();
+    }: { call.dispatch_bypass_filter(origin)? }
```
kaichaosun, line 72: Most of the time to call withdraw or slash, we aren't using root, but a member of council.

pallets/liquid-staking/src/benchmarking.rs
```diff
@@ -0,0 +1,251 @@
...
+        );
+        assert_eq!(
+            <T as Config>::Currency::free_balance(T::StakingCurrency::get(), &LiquidStaking::<T>::account_id()),
+            INITIAL_AMOUNT + amount
+        );
+
+        assert_last_event::<T>(Event::<T>::Unstaked(caller, unstake_amount, asset_amount).into());
+    }
+
+    process_pending_unstake {
```
kaichaosun, line 152: `process_pending_unstake` and `finish_processed_unstake ` need to take consideration of `MaxAccountProcessingUnstake `.  The worst scenario is the max number of this constant, but it's possible to use this constant as a variable in the weight annotation.

pallets/liquid-staking/src/benchmarking.rs
```diff
@@ -0,0 +1,251 @@
...
+        );
+        assert_eq!(
+            <T as Config>::Currency::free_balance(T::LiquidCurrency::get(), &caller),
+            5_000_000
+        );
+        assert_eq!(
+            <T as Config>::Currency::free_balance(T::StakingCurrency::get(), &LiquidStaking::<T>::account_id()),
+            INITIAL_AMOUNT + amount
+        );
+        assert_last_event::<T>(Event::<T>::Staked(caller, amount).into());
```
ayushmishra2005, line 58: Asserting Last event would be enough.

pallets/liquid-staking/src/benchmarking.rs
ayushmishra2005: As discussed, I will be using `T::WithdrawOrigin::successful_origin();`

pallets/liquid-staking/Cargo.toml
```diff
@@ -29,8 +30,10 @@ sp-io           = { git = 'https://github.com/paritytech/substrate.git', branch
 
 [features]
 default = ['std']
+runtime-benchmarks = ["frame-benchmarking"]
```
GopherJ, line 33: let's use single quote here

.github/workflows/coverage.yml
```diff
@@ -68,7 +68,7 @@ jobs:
 
       - name: Generate code coverage
         run: |
-          cargo +nightly-2021-03-24 tarpaulin --verbose --all-features --workspace --exclude parallel --exclude parallel-dev --exclude parallel-runtime --exclude vanilla-runtime --exclude heiko-runtime --exclude pallet-loans-benchmarking  --exclude-files pallets/liquid-staking/src/benchmarking.rs --timeout 120 --out Xml
+          cargo +nightly-2021-03-24 tarpaulin --verbose --all-features --workspace --exclude parallel --exclude parallel-dev --exclude parallel-runtime --exclude vanilla-runtime --exclude heiko-runtime --exclude pallet-loans-benchmarking  --exclude pallet-liquid-staking --timeout 120 --out Xml
```
GopherJ, line 71: why did you exclude the whole crate? we still need to count coverage of liquid staking pallet, but we just need to ignore the benchmarking part

GopherJ (changes requested):

.github/workflows/coverage.yml
ayushmishra2005: `--exclude-files` is not working. I just wanted to check whether this change would work or not

kaichaosun: It's fine to exclude this crate for now, please create a issue to track the context so that we can add it back later.

kaichaosun (approved):
👍 Looks great.

GopherJ (approved):

PR #247, opened 2021-06-25
New curve model

c410-f3r:
## `x` and `y` axis (Domains)

By looking at https://docs.parallel.fi/white-paper#3-experimental-models, it is possible to see on the Cartesian plane that both axis are 0 to 1 real numbers that represent percentages, therefore, a equation for the new curve has `x ∈ ℝ | 0 <= x <= 1` and `y ∈ ℝ | 0 <= x <= 1` domains.

## Curve choice

It is still not clear to me yet which curve should perform best for Parallel. For now, any exponential formula with an adjustable constant should be good enough for experimentation because if the output value is too high or too low, then the constant could be easily modified.
Taking into consideration the above considerations, it is possible to use `y = (C^x - 1)/(C - 1), where C = constant`.

<sub>y = (256^x - 1)/(256 - 1)</sub>
save

Or `y = x^9 + C, where C = base_rate and y > 0` as suggested by @yz89.

<sub>y = x^9 + 0.02</sub>
save(1)

## Other considerations

* The equation of https://docs.parallel.fi/white-paper#3-1-curve-interest-rate-model does not specify what the `b` and `c` variables mean or the range of the `n` variable used for summing. Due to this situation and by request, another equation was used to evaluate `BorrowInterestRate` but future formula modifications can still occur if necessary.
* This PR only adds the new curve using `y = x^9 + C` for calculation. Future PRs can be created to actually change the internal algebraic operations of loans with this model.

### DIFF
--- a/pallets/loans/src/rate_model.rs
+++ b/pallets/loans/src/rate_model.rs
@@ -175,9 +175,10 @@ impl CurveModel {
 
     /// Calculates the borrow interest rate of curve model
     pub fn get_borrow_rate(&self, utilization: Ratio) -> Option<Rate> {
+        const NINE: usize = 9;
         let utilization_rate: Rate = utilization.into();
         utilization_rate
-            .saturating_pow(9)
+            .saturating_pow(NINE)
             .checked_add(&self.base_rate)
     }
 }
--- a/pallets/loans/src/rate_model.rs
+++ b/pallets/loans/src/rate_model.rs
@@ -161,6 +161,8 @@ pub struct CurveModel {
 }
 
 impl CurveModel {
+    pub const MAX_BASE_RATE: Rate = Rate::from_inner(Rate::DIV / 100 * 10); // 10%
+
     /// Create a new curve model
     pub fn new_model(base_rate: Rate) -> CurveModel {
         Self { base_rate }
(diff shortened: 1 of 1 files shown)


pallets/loans/src/rate_model.rs
```diff
@@ -12,16 +12,26 @@
...
+    fn get_borrow_rate(&self, utilization: Ratio) -> Option<Rate>;
+}
 
 /// Parallel interest rate model
 #[derive(Encode, Decode, Eq, PartialEq, Copy, Clone, RuntimeDebug)]
 pub enum InterestRateModel {
-    Jump(JumpModel),
     Curve(CurveModel),
+    Jump(JumpModel),
+    Polynomial(PolynomialModel),
```
yz89, line 34: PolynomialModel is still under discussion and should not add to this PR. Please remove this related code and let this PR be merged asap.

pallets/loans/src/lib.rs
```diff
@@ -89,6 +89,9 @@ pub mod pallet {
     pub trait Config: frame_system::Config {
         type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
 
+        /// Borrow rate used to calculate polynomial rate models.
+        type BorrowRate: Get<Rate>;
```
yz89, line 93: BorrowRate has no usage in this PR.

c410-f3r: OK. Taking into considerations the feedback, this PR now does not depend on #231 anymore.

pallets/loans/src/rate_model.rs
```diff
@@ -161,20 +161,24 @@ pub struct CurveModel {
...
     }
 
     /// Calculates the borrow interest rate of curve model
-    pub fn get_borrow_rate(&self, _utilization: Ratio) -> Option<Rate> {
-        // TODO: Need to implement the curve model
-        None
+    pub fn get_borrow_rate(&self, utilization: Ratio) -> Option<Rate> {
+        let utilization_rate: Rate = utilization.into();
+        utilization_rate
+            .saturating_pow(9)
```
yz89, line 180: 9 could be a const var.

yz89 (approved):

ayushmishra2005 (approved):

yz89 (approved):

Issue #244, opened 2021-06-25
redeem_all doesn't really redeem all tokens


GopherJ:

GopherJ:
<img width="1369" alt="image" src="https://user-images.githubusercontent.com/33961674/123353116-6406ad00-d593-11eb-8867-0b43d2fe7558.png">

GopherJ:
https://github.com/parallel-finance/parallel-dapp/issues/243

GopherJ:
the problem it caused is that the collateral button is always green, we need to find a way to fix it

PR #249, opened 2021-06-26
fix precision loss in redeem_all

GopherJ:
close: #244 
close: https://github.com/parallel-finance/parallel-dapp/issues/243

### DIFF
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -1368,7 +1368,7 @@ impl<T: Config> Pallet<T> {
     ) -> Result<Balance, DispatchError> {
         Ok(exchange_rate
             .checked_mul_int(voucher_amount)
-            .ok_or(ArithmeticError::Underflow)?)
+            .ok_or(ArithmeticError::Overflow)?)
     }
 
     pub fn calc_collateral_amount(
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -505,7 +505,7 @@ pub mod pallet {
             let exchange_rate = Self::exchange_rate(currency_id);
             let voucher_amount = Self::calc_collateral_amount(redeem_amount, exchange_rate)?;
             Self::update_earned_stored(&who, &currency_id)?;
-            Self::redeem_internal(&who, &currency_id, voucher_amount)?;
+            let redeem_amount = Self::redeem_internal(&who, &currency_id, voucher_amount)?;
 
             Self::deposit_event(Event::<T>::Redeemed(who, currency_id, redeem_amount));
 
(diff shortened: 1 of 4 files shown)


pallets/loans/src/lib.rs
```diff
@@ -1366,15 +1362,23 @@ impl<T: Config> Pallet<T> {
         T::Currency::free_balance(currency_id, &Self::account_id())
     }
 
+    pub fn calc_underlying_amount(
+        voucher_amount: u128,
+        exchange_rate: Rate,
+    ) -> Result<Balance, DispatchError> {
+        Ok(exchange_rate
+            .checked_mul_int(voucher_amount)
+            .ok_or(ArithmeticError::Underflow)?)
```
yz89, line 1371: Should be ArithmeticError:Overflow

pallets/loans/src/lib.rs
```diff
@@ -916,16 +914,16 @@ impl<T: Config> Pallet<T> {
     fn redeem_allowed(
         currency_id: &CurrencyId,
         redeemer: &T::AccountId,
-        redeem_amount: Balance,
-    ) -> DispatchResult {
+        voucher_amount: Balance,
+    ) -> Result<Balance, DispatchError> {
```
yz89, line 918: It is odd to return a balance here. I suggest putting this check into `redeem` and `redeem_all` separately.

GopherJ: I think it's ok because `ensure_origin` also returns back the AccountId:

> I suggest putting this check into redeem and redeem_all separately.

yz89: To be consistent with other methods, such as Borrow_Allowed, there may be liquidate_allowed. This is just to see if you can continue.

GopherJ: do you have an idea on how should I write it?

pallets/loans/src/lib.rs
```diff
@@ -916,16 +914,16 @@ impl<T: Config> Pallet<T> {
...
         let deposit = Self::account_deposits(currency_id, redeemer);
         let exchange_rate = Self::exchange_rate(currency_id);
-        let voucher_amount = Self::calc_collateral_amount(redeem_amount, exchange_rate)?;
+        let redeem_amount = Self::calc_underlying_amount(voucher_amount, exchange_rate)?;
```
yz89, line 921: redeem_amount has no usage in this function.

GopherJ: it's used in `redeem_internal`

yz89: That should be calculated in `redeem_internal`.

ayushmishra2005 (approved):

pallets/loans/src/lib.rs
GopherJ: @yz89 as you can see from here: https://github.com/parallel-finance/parallel/pull/249/commits/a5f7ee0fba725cd5315ed5b248c5f9f7897641bd

if I remove the return value, there will be some unecessary calculations

yz89 (approved):

PR #219, opened 2021-06-16
fix the issue of not accrue interest

yz89:
closes #201

PR #250, opened 2021-06-26
add price ticker

yz89:


GopherJ (approved):

GopherJ:
so I should subscribe to Orml oracle's storage right?

GopherJ:
it seems that oracle's storage wasn't updated
image

yz89:
> so I should subscribe to Orml oracle's storage right?

Yes, or you can query the RPC in time.

pallets/loans/src/lib.rs
```diff
@@ -446,6 +446,11 @@ pub mod pallet {
                     }
                 }
             });
+
+            // This is used to trigger the price aggregation to update the results to the ORML Oracle Pallet.
+            for currency_id in Currencies::<T>::get().iter() {
+                let _ = Self::get_price(currency_id);
```
kaichaosun, line 452: This needs to be better weighted.

yz89: Is any suggestion here?

kaichaosun: The `on_initialized` function should be able to return the weight.

yz89: We moved the code into `on_finalize` in #219 and maybe needn't anymore.

Issue #252, opened 2021-06-26
repay_borrow_all underflow


GopherJ:
<img width="1104" alt="image" src="https://user-images.githubusercontent.com/33961674/123517085-da112e00-d6d1-11eb-85a0-d3e8a99a580c.png">

<img width="517" alt="image" src="https://user-images.githubusercontent.com/33961674/123517101-e2696900-d6d1-11eb-9697-b0aa29f178ad.png">

GopherJ:
<img width="330" alt="image" src="https://user-images.githubusercontent.com/33961674/123517119-f1e8b200-d6d1-11eb-88d4-20519ed78f0d.png">

GopherJ:
https://github.com/parallel-finance/parallel-dapp/issues/258

yz89:
Is this issue fixed?

GopherJ:
@yz89 it's fixed I didn't close it because I'm wondering if there are better ideas

yz89:
Got it

PR #254, opened 2021-06-27
fix repay_borrow_all

GopherJ:


pallets/loans/src/lib.rs
```diff
@@ -1015,10 +1016,10 @@ impl<T: Config> Pallet<T> {
             .checked_sub(repay_amount)
             .ok_or(ArithmeticError::Underflow)?;
         let total_borrows = Self::total_borrows(currency_id);
-        let total_borrows_new = total_borrows
-            .checked_sub(repay_amount)
-            .ok_or(ArithmeticError::Underflow)?;
+        // TODO : total_borrows should contain interests
+        let total_borrows_new = total_borrows.saturating_sub(repay_amount);
```
GopherJ, line 1020: total_borrows < repay_amount and caused this issue:)

we need to find a way to make total_borrows contain also interests

GopherJ, line 1020: total_borrows < repay_amount and caused this issue:)

we need to find a way to make total_borrows contain also interests

yz89: I think we can use this PR to fix #252. It only appears in the early days when total_borrows is too small.
When the case appears, the result of this modification is that `exchange_rate` won't be updated and no lender will get supply interest. That shouldn't be a problem.

yz89: Could you modify the TODO to a NOTE and explain the code briefly? And pls add an edge test to catch this case.

yz89:
I think we can use this PR after addressed the comments.

pallets/loans/src/lib.rs
GopherJ: ok I'll do it

pallets/loans/src/lib.rs
```diff
@@ -1035,17 +1036,17 @@ impl<T: Config> Pallet<T> {
 
     // Calculates and returns the most recent amount of borrowed balance of `currency_id`
     // for `who`.
-    pub fn borrow_balance_stored(
+    pub fn recent_borrow_balance(
```
kaichaosun, line 1039: The name refactor doesn't make sense, `recent` is too general.

GopherJ: do you have an suggestion here? like `latest`

yz89: Maybe `current`

GopherJ:
@yz89  pls check latest commit, it'll catch this error if we change back to `checked_..`

yz89 (approved):

PR #255, opened 2021-06-27
fetch liquid price and add test

MrPai:
Now the price of xKSM will calculate according to KSM and exchangeRate. I haven't delete xKSM in Oracle-client since the frontend will get price directly from `oracle pallet`.

### DIFF
--- a/pallets/prices/src/lib.rs
+++ b/pallets/prices/src/lib.rs
@@ -25,10 +25,15 @@
 
 use frame_support::{pallet_prelude::*, transactional};
 use frame_system::pallet_prelude::*;
+use orml_oracle::DataProviderExtended;
 use orml_traits::DataProvider;
 pub use pallet::*;
 use primitives::*;
-use sp_runtime::{traits::CheckedDiv, FixedU128};
+use sp_runtime::{
+    traits::{CheckedDiv, CheckedMul},
+    FixedU128,
+};
+use sp_std::vec::Vec;
 
 #[cfg(test)]
 mod mock;
@@ -44,10 +49,23 @@ pub mod pallet {
         type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
 
         /// The data source, such as Oracle.
-        type Source: DataProvider<CurrencyId, TimeStampedPrice>;
+        type Source: DataProvider<CurrencyId, TimeStampedPrice>
+            + DataProviderExtended<CurrencyId, TimeStampedPrice>;
 
         /// The origin which may set prices feed to system.
         type FeederOrigin: EnsureOrigin<Self::Origin>;
+
+        /// Currency used for staking
+        #[pallet::constant]
+        type StakingCurrency: Get<CurrencyId>;
+
+        /// Currency used for liquid voucher
+        #[pallet::constant]
+        type LiquidCurrency: Get<CurrencyId>;
+
+        /// The provider of the exchange rate between liquid currency and
+        /// staking currency.
+        type LiquidStakingExchangeRateProvider: ExchangeRateProvider;
     }
 
     #[pallet::event]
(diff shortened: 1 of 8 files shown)


GopherJ:
will frontend get different price than backend? @MrPai

MrPai:
I think frontend won't get xKSM price from `oracle pallet` if we remove the xKSM in oracle client.

GopherJ:
> I think frontend won't get xKSM price from `oracle pallet` if we remove the xKSM in oracle client. @GopherJ

shall frontend do the same calculation using exchange rate?

MrPai:
I think the backend supposed to have xKSM price calculated on-chain, because while we regard xKSM as collateral, the `loans pallet` should get xKSM price from `price pallet`.
Although we can calculate xKSM in frontend but I don't suggest it. let's see if we can get xKSM from `price pallet`.
btw, here is the Acala's implement about `get_price()` that 
 I referred to: https://github.com/AcalaNetwork/Acala/blob/master/modules/prices/src/lib.rs#L167

ayushmishra2005 (approved):

GopherJ:
I'll finish the frontend changes and put the pr here so that you guys can help reviewing

GopherJ (approved):

runtime/vanilla/src/lib.rs
```diff
@@ -566,17 +566,37 @@ impl pallet_loans::Config for Runtime {
...
 pub struct AggregatedDataProvider;
 impl DataProvider<CurrencyId, TimeStampedPrice> for AggregatedDataProvider {
     fn get(key: &CurrencyId) -> Option<TimeStampedPrice> {
         Oracle::get(key)
     }
 }
+
+impl DataProviderExtended<CurrencyId, TimeStampedPrice> for AggregatedDataProvider {
+    fn get_no_op(key: &CurrencyId) -> Option<TimeStampedPrice> {
+        Oracle::get_no_op(key)
```
yz89, line 578: Shouldn't this is Prices::get_no_op(key)?

MrPai: It's Oracle actually, the method you point out in runtime will be invoked by `T::Source::get_no_op()` in the Price pallet.

yz89: Ok, I see.

pallets/prices/src/lib.rs
```diff
@@ -124,14 +142,30 @@ impl<T: Config> PriceFeeder for Pallet<T> {
...
-                    p.value
-                        .price
-                        .checked_div(&FixedU128::from_inner(d))
-                        .map(|price| (price, p.timestamp))
+            if currency_id == &T::LiquidCurrency::get() {
+                T::Source::get(&T::StakingCurrency::get()).and_then(|p| {
+                    10u128.checked_pow(p.value.decimal.into()).and_then(|d| {
+                        p.value
+                            .price
+                            .checked_div(&FixedU128::from_inner(d))
```
kaichaosun, line 150: You already have decimal in prices info, why still use from_inner here?

runtime/vanilla/src/lib.rs
```diff
@@ -566,17 +566,37 @@ impl pallet_loans::Config for Runtime {
     type UnixTime = Timestamp;
 }
 
-pub type TimeStampedPrice = orml_oracle::TimestampedValue<PriceWithDecimal, Moment>;
 pub struct AggregatedDataProvider;
```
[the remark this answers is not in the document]
MrPai: `AggregatedDataProvider` implement `DataProvider&DataProviderExtended` with specific type, however the `orml-oracle` pallet implement it with generic types, pls check here: https://github.com/open-web3-stack/open-runtime-module-library/blob/master/oracle/src/lib.rs#L266

runtime/heiko/src/lib.rs
```diff
@@ -649,10 +649,30 @@ impl DataProvider<CurrencyId, TimeStampedPrice> for AggregatedDataProvider {
...
     type Event = Event;
     type Source = AggregatedDataProvider;
     type FeederOrigin = EnsureRoot<AccountId>;
+    type StakingCurrency = StakingCurrency;
+    type LiquidCurrency = LiquidCurrency;
+    type LiquidStakingExchangeRateProvider = LiquidStakingExchangeRateProvider;
+}
+
+pub struct LiquidStakingExchangeRateProvider;
+impl ExchangeRateProvider for LiquidStakingExchangeRateProvider {
```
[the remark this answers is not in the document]
MrPai: I think that injects a single struct here is more decoupling and effective instead of a ‘big’ pallet, because `LiquidStaking` contains more than an exchange rate. how do you think of it? :)

pallets/prices/src/lib.rs
MrPai: this is the only place that we care about decimal on-chain, while other pallet query prices from this `price pallet`, will return a minimum unit of currency price, so we need to divide the decimal.

MrPai:
@kaichaosun I remove struct `LiquidStakingExchangeRateProvider` and implement the logic in liquid-staking, please review if we merge this?

GopherJ:
@MrPai pls check: https://github.com/parallel-finance/parallel-dapp/pull/269

PR #271, opened 2021-07-05
Improve/parachain workflow

GopherJ:
close: #266 
close: #272


Dockerfile.vanilla
```diff
@@ -0,0 +1,30 @@
+FROM paritytech/ci-linux:production as builder
```
MrPai, line 1: Can we merge these two Dockerfiles into one?

MrPai (commented):
`root_key` for  heiko-dev chain spec should be common account, such as Alice;
and there is no need to include "Parallel team accounts" in heiko-dev

Dockerfile.vanilla
GopherJ: the default command isn't the same

GopherJ:
the others are ok now

Makefile
```diff
@@ -79,14 +79,19 @@ resources:
 	target/release/parallel export-genesis-state --parachain-id 200 > ./resources/para-200-genesis
 	target/release/parallel export-genesis-wasm > ./resources/para-200.wasm
 
-.PHONY: register-parachain
-register-parachain:
-	polkadot-js-api --seed "//Alice" --sudo tx.parasSudoWrapper.sudoScheduleParaInitialize 200 "{ \"genesisHead\":\"$(shell cat ./resources/para-200-genesis)\", \"validationCode\":\"$(shell cat ./resources/para-200.wasm)\", \"parachain\": true }"
+.PHONY: docker-resources
+docker-resources:
+	docker run --rm  parallelfinance/parallel:latest parallel export-genesis-state --chain heiko-dev --parachain-id 200 > ./resources/para-200-genesis
```
MrPai, line 84: parachain-id should start from 2000, you can refer to https://substrate.dev/cumulus-workshop/#/en/2-relay-chain/2-reserve?id=reserve-a-paraid

GopherJ: we haven't decided our paraid yet so 200 is only for testing

GopherJ:
ok now it starts to produce blocks

MrPai (approved):
looks good to me

PR #286, opened 2021-07-12
Improve liquid staking

MrPai:
add approved agent list and add relaychain era_index

### DIFF
--- a/pallets/liquid-staking/src/benchmarking.rs
+++ b/pallets/liquid-staking/src/benchmarking.rs
@@ -64,11 +64,12 @@ where
         SystemOrigin::Signed(caller.clone()).into(),
         unstake_amount
     ));
-
+    let era_index = 1;
     assert_ok!(LiquidStaking::<T>::process_pending_unstake(
         T::WithdrawOrigin::successful_origin(),
         agent,
         caller,
+        era_index,
         amount
     ));
 }
@@ -183,7 +184,8 @@ benchmarks! {
         assert_ok!(LiquidStaking::<T>::unstake(
             SystemOrigin::Signed(caller.clone()).into(),
             unstake_amount));
-        let call = Call::<T>::process_pending_unstake(agent.clone(), caller.clone(), amount);
+        let era_index = 1;
+        let call = Call::<T>::process_pending_unstake(agent.clone(), caller.clone(), era_index, amount);
         let origin = T::WithdrawOrigin::successful_origin();
     }: { call.dispatch_bypass_filter(origin)? }
     verify {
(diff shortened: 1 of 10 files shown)


runtime/parallel/src/lib.rs
```diff
@@ -361,6 +361,24 @@ impl pallet_loans::Config for Runtime {
...
 }
 
+parameter_types! {
+    pub const LiquidStakingAgentMaxMembers: u32 = 100;
+}
+
+type LiquidStakingAgentMembershipInstance = pallet_membership::Instance3;
+impl pallet_membership::Config<LiquidStakingAgentMembershipInstance> for Runtime {
+    type Event = Event;
+    type AddOrigin = EnsureRoot<AccountId>;
```
GopherJ, line 371: we should also allow council to update

GopherJ:
otherwise it looks fine to me

RainFallsSilent:
I think this pull request should be merged after stake-client is ready to deploy.

GopherJ:
also @MrPai do we have some hooks to use in membership pallet? such as `onRemove` we need hooks to handle the tokens left in agent account.

MrPai:
> also @MrPai do we have some hooks to use in membership pallet? such as `onRemove` we need hooks to handle the tokens left in agent account.

None for now, the agent is a multi-sig account on relaychain, we should confirm that there is no token bonded by it before we remove.  however "bond action" is on relaychain, but "onRemove" is hooked on parachain, there may be some details to design.

I agree this is necessary but should have a clear process before.

MrPai:
> I think this pull request should be merged after stake-client is ready to deploy.

Actually we can merge it now, because current `stake-client` testing is based on a docker-composer file that in another branch, don't worry about the onchain interaction with stake-client.

pallets/liquid-staking/src/benchmarking.rs
```diff
@@ -64,11 +64,12 @@ where
         SystemOrigin::Signed(caller.clone()).into(),
         unstake_amount
     ));
-
+    let era_index = 1;
```
kaichaosun, line 67: There is no need for this assignment, just use `1` in assert_ok should be fine.

pallets/liquid-staking/src/lib.rs
```diff
@@ -483,6 +501,7 @@ pub mod pallet {
             let new_unstake = UnstakeInfo {
                 amount,
                 block_number,
+                era_index: Some(era_index),
```
kaichaosun, line 504: Era index should be updated after the offchain staking client process it successfully.

MrPai: As I know, the Era index should be fixed after we unbond on relaychain. if we unbond a second time on relaychain, it is totally another Era index.
The era index is associated with the amount each time invoking `unbond`.
please correct me if I have any misunderstandings.

kaichaosun (approved):
As the `process_pending_unstake` is asynchronous, need a way to identify if it's success or not.

PR #292, opened 2021-07-12
change native token to HKO/PARA

GopherJ:
frontend is implmenting tx fee, so we need to add new tokens: PARA/HKO

companion pr: https://github.com/parallel-finance/parallel-dapp/pull/307

### DIFF
--- a/runtime/vanilla/src/lib.rs
+++ b/runtime/vanilla/src/lib.rs
@@ -519,7 +519,7 @@ impl orml_tokens::Config for Runtime {
 }
 
 parameter_types! {
-    pub const GetNativeCurrencyId: CurrencyId = CurrencyId::Native;
+    pub const GetNativeCurrencyId: CurrencyId = CurrencyId::HKO;
 }
 impl orml_currencies::Config for Runtime {
     type Event = Event;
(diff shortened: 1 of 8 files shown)


GopherJ:
we shouldn't merge this util it's tested and the `PARA/HKO` token's icon has been designed

runtime/vanilla/src/lib.rs
```diff
@@ -519,7 +519,7 @@ impl orml_tokens::Config for Runtime {
 }
 
 parameter_types! {
-    pub const GetNativeCurrencyId: CurrencyId = CurrencyId::Native;
+    pub const GetNativeCurrencyId: CurrencyId = CurrencyId::HKO;
```
kaichaosun, line 522: This should be something like `DEV`.

kaichaosun (approved):

crystalynliu:
After the PR merged, we should also change the decimal and token name with configuration.

https://github.com/parallel-finance/parallel-dapp/blob/testnet/src/transaction/index.js#L59
https://github.com/parallel-finance/parallel-dapp/blob/testnet/src/pages/LendAndBorrow/Transaction/index.js#L156

runtime/vanilla/src/lib.rs
GopherJ: @yz89 what do we want exactly?

GopherJ: @yz89 do we really need a currency for vanilla?

GopherJ: looks like acala doesn't have a currency for mandala https://github.com/AcalaNetwork/Acala/blob/master/primitives/src/currency.rs#L175-L193 @yz89 @kaichaosun

yz89: Ok, I now think it make sense to use `HKO`. It shouldn't to take a place on CurrencyId.

PR #303, opened 2021-07-13
AMM MVP

c410-f3r:
## Preamble

This PR is not meant be a final/deliverable/finished product but rather a result summary of my brief research on a potential suitable curve for Parallel's AMM. Since I will be working on a different task now, it is possible to archive everything that was written after any desired inspection for future references.

It is also worth noting that this research is focused on `1:1` asset ratios (`TUSD`/`USDT`, `KSM`/`xKSM`, etc). If appropriated, heterogeneous ratio pools can be further investigated in the future.

## The problem

Constant product based AMMs (`x * y = C`) are known for their high slippage when considerable amounts of assets are swapped. For example, a pool with initial assets `x(0) = 50` and `y(0) = 50` have a price `p(x) = 1` while the same pool with assets `x(n) = 25` and `y(n) = 75` would have a price `p(x) = 4` totalling a 400% difference.

## One possible solution

On the opposite side of `x * y = C`, a constant sum equation (`x + y = C`) always have `p(x) = 1` independently of `x` but lacks liquidity so a middle ground between the two formulas could decrease slippage and provide a respectable amount of liquidity at the same. That is exactly what Curve developed and deployed on the Ethereum network.

Capture d’écran de 2021-07-12 22-29-44

As the point on the curve moves towards the equilibrium point (`x = y`), the graph resembles `x + y` while the opposite causes the graph to act like `x * y`.

## This PR

Instead of creating a new curve (would take more time), a simple pallet that stores pools and performs swaps was developed using the function created by the Curve team simplified for 2 assets instead of N.

```rust
let A = ADJUSTABLE CONSTANT;
let C = ASSETS QUANTITY;

let y = (C * (4*A*C + C - 4*A*x)) / (4 * (A*C + x))
```

No runtime integration was performed and some tests were added as a functional proof-of-concept.

## Sources

https://arxiv.org/pdf/2101.02778.pdf
https://curve.fi/files/stableswap-paper.pdf
https://docs.pendle.finance/getting-started/pendle-amm
https://medium.com/bollinger-investment-group/constant-function-market-makers-defis-zero-to-one-innovation-968f77022159
https://yield.is/yieldspace.pdf


c410-f3r:
The curve is now pegged by the exchange rate plus an additional weight.
An exchange rate of 0% is more geared towards an uniswap-like formula while a 100% exchange rate moves down to the constant sum equation. In other words, the larger the exchange rate, the more invariant is the asset price.

Capture d’écran de 2021-07-17 23-47-22

Code was re-based, re-formulated and more documentation was added.

pallets/amm/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        // Sum of the two pool assets
+        fn total_assets(pool: &Pool) -> Result<Balance, DispatchError> {
+            pool.amount_base
+                .checked_add(pool.amount_quote)
+                .ok_or_else(|| ArithmeticError::Overflow.into())
+        }
+    }
+
+    #[pallet::storage]
+    pub type ExchangeRate<T: Config> = StorageValue<_, Rate, ValueQuery>;
```
c410-f3r, line 226: Just to show how the exchange rate would be used to shape the curve.

Both `amm` and `loans` use stored `exchange_rate` values so it is probably better to created a common pallet instead of duplicating error-probe code.

yubo-ruan:
We thought about the end goal of the new AMM curve. We should add a test to the current formula and make sure it achieves our design requirement. 
- I think we should do some testing on the formula first. It doesn't have to be using Substrate and Rust (what about Python or Matlab?)
- After this, we should hop on a call to discuss more the Substrate implementation.  

**Regarding the testing requirement** 
First, we need to pick a no-slippage range number. Let's say we choose 1/6. 
Second:
- When users sell less than 1/6 of the pool size, we shouldn't incur any slippage since users are swapping on the constant sum portion
- When users sell more than 1/6 of the pool size, users should incur slippage, and slippage should increase fast enough to provide an incentive for the counter trade
- the above requirement should always be met given any arbitrary exchange rate that is above 1 (stimulate the exchange rate in the next 5 years)

c410-f3r:
> We thought about the end goal of the new AMM curve. We should add a test to the current formula and make sure it achieves our design requirement.
> 
>     * I think we should do some testing on the formula first. It doesn't have to be using Substrate and Rust (what about Python or Matlab?)
> 
>     * After this, we should hop on a call to discuss more the Substrate implementation.
> 
> 
> **Regarding the testing requirement**
> First, we need to pick a no-slippage range number. Let's say we choose 1/6.
> Second:
> 
>     * When users sell less than 1/6 of the pool size, we shouldn't incur any slippage since users are swapping on the constant sum portion
> 
>     * When users sell more than 1/6 of the pool size, users should incur slippage, and slippage should increase fast enough to provide an incentive for the counter trade
> 
>     * the above requirement should always be met given any arbitrary exchange rate that is above 1 (stimulate the exchange rate in the next 5 years)

Thank you for the feedback @yubo-ruan. Everything shall be properly addressed ASAP

c410-f3r:
Some rust code was created to discover a coefficient parameter that would satisfy all required constraints and after some empirical tests, a coefficient weight of 0.5 was found for a curve without slippage for amounts up to 12% of the total pool, an exchange rate of 1 (100%)  and a maximum slippage tolerance of 150% (worse-case scenario when all assets are swapped).

amm-formula

Above is the graph generated by the commented rust code located at https://github.com/parallel-finance/amm-formula and two unit tests were included to assert slippage percentages (https://github.com/parallel-finance/parallel/pull/303/files#diff-1c3d96d1ce350c4da5648877c7735536ff17f5ed90a4c096e2b947862aaa03e6R39-R81).

No proof was formulated. Every value was retrieved in a empirical manner with the following features:

* The higher the exchange rate, the more invariant is the curve.
* The more the slippage tolerance, the more difficult is to get high slippages that are out of the tolerance range.

If higher slippage tolerance and a higher slippage percentage is desired, then I recommend using a conditional curve composed by the constant sum formula and the curve formula, i.e., if `x <= max slippage tolerance`, then use constant sum; if `x > max slippage tolerance`, then use the curve formula with a desired slippage. Example:

example

yz89:
Hi @c410-f3r, Thanks for the excellent work. 
As this PR is experimental, I will close it for now and we'll talk about that when we actually going to implement the insurance pool and stability pool.

PR #321, opened 2021-07-22
Feature/ss58 format

alannotnerd:
Add system properties for parallel and heiko, but left it default for local test network. 

ss58 address format id is registered in paritytech/substrate#9254


node/parallel/src/chain_spec/mod.rs
```diff
@@ -15,13 +15,26 @@
...
 use sp_consensus_aura::sr25519::AuthorityId as AuraId;
 use sp_core::{sr25519, Pair, Public};
 use sp_runtime::traits::IdentifyAccount;
 
+use primitives::{network::NetworkType, *};
+
+pub(crate) fn as_properties(network: NetworkType) -> Properties {
+    json!({
+        "ss58Format": network.ss58_addr_format_id(),
+        "network": network,
```
GopherJ, line 31: hi @alannotnerd  where will this `network` field be used?

alannotnerd: It effects the display information in polkadotjs dapps, just like `tokenSymbol` field.

alannotnerd: Sorry, I checked the polkadotjs source code just now and found it's not the standard field in properties. I was thounght it's the name displayed in upper left corner. But in fact, it's from `api.system.name()`. I'll remove it later. BTW, what are the token name and its decimal? I can add them in the next commit incidentally.

GopherJ: > Sorry, I checked the polkadotjs source code just now and found it's not the standard field in properties. I was thounght it's the name displayed in upper left corner. But in fact, it's from `api.system.name()`. I'll remove it later. BTW, what are the token name and its decimal? I can add them in the next commit incidentally.

no problem, we have `HKO` for heiko network and `PARA` for parallel network, they all have 12 as token decimals

GopherJ:
@alannotnerd hi I think acala has given a good example regarding the token decimals

alannotnerd:
The properties just declares our token's decimals.

node/parallel/src/chain_spec/mod.rs
```diff
@@ -15,13 +15,31 @@
...
+/// For fields definition, see https://github.com/polkadot-js/apps/blob/bd78840d2142df121d182e8700b20308880dde0a/packages/react-api/src/Api.tsx#L115
+pub(crate) fn as_properties(network: NetworkType) -> Properties {
+    let (symbol, decimal) = network.token_info();
+    json!({
+        "ss58Format": network.ss58_addr_format_id(),
+        "tokenSymbol": symbol,
+        "tokenDecimals": decimal,
+    })
+    .as_object()
+    .expect("Network properties are valid; qed")
```
kaichaosun, line 39:
```suggestion
    .expect("Network properties are invalid; qed")
```

node/parallel/src/chain_spec/heiko.rs
```diff
@@ -74,8 +76,8 @@ pub fn development_config(id: ParaId) -> ChainSpec {
         },
         vec![],
         None,
-        None,
-        None,
+        Some("dev"),
```
kaichaosun, line 79: Heiko is not a dev chain.

node/parallel/src/chain_spec/heiko.rs
```diff
@@ -161,8 +163,8 @@ pub fn local_testnet_config(id: ParaId) -> ChainSpec {
...
```
kaichaosun, line 166: Same here.

node/parallel/src/chain_spec/parallel.rs
```diff
@@ -73,8 +75,8 @@ pub fn development_config(id: ParaId) -> ChainSpec {
         },
         vec![],
         None,
-        None,
-        None,
+        Some("dev"),
```
kaichaosun, line 78: Same.

node/parallel/src/chain_spec/parallel.rs
```diff
@@ -160,8 +162,8 @@ pub fn local_testnet_config(id: ParaId) -> ChainSpec {
...
```
kaichaosun, line 165: And here.

primitives/src/network.rs
```diff
@@ -0,0 +1,37 @@
+use codec::{Decode, Encode};
+#[cfg(feature = "std")]
+use serde::{Deserialize, Serialize};
+use sp_runtime::RuntimeDebug;
+
+/// Network type for parallel.
+#[derive(PartialEq, Eq, Clone, Copy, Encode, Decode, RuntimeDebug)]
```
kaichaosun, line 7: Encode, decode, Eq, should not be needed, only add the derived implementation that needed.

primitives/src/network.rs
```diff
@@ -0,0 +1,37 @@
...
+pub enum NetworkType {
+    Parallel,
+    Heiko,
+}
+
+impl NetworkType {
+    /// Return ss58 address prefix from network type.
+    pub fn ss58_addr_format_id(&self) -> u8 {
+        match self {
+            NetworkType::Heiko => 110,
```
kaichaosun, line 18: Use const definition.

primitives/src/network.rs
```diff
@@ -0,0 +1,37 @@
...
+            NetworkType::Parallel => 172,
+        }
+    }
+
+    /// Return (token_symbol, token_decimal) of this network.
+    pub fn token_info(&self) -> (&str, u8) {
+        match self {
+            NetworkType::Heiko => (HEIKO_TOKEN, 12),
```
kaichaosun, line 26: The token decimals should not put in the chain_spec, not the primitives, so that won't miss use it. How about put the network source near to chain_spec (in same directory).

node/parallel/src/chain_spec/heiko.rs
alannotnerd: Ok, I know, but it's in development config not in live conifg. Do we share the config in both network?

node/parallel/src/chain_spec/mod.rs
alannotnerd: I think it proves why the statement won't be panic.  See https://github.com/paritytech/substrate/blob/1d5abf01abafdb6c15bcd0172f5de09fd87c5fbf/primitives/npos-elections/src/balancing.rs#L122

primitives/src/network.rs
alannotnerd: Fix it in next commit.

primitives/src/network.rs
alannotnerd: Fix it in next commit.

primitives/src/network.rs
alannotnerd: Of course it's the better solution if we only use it in chain_spec, but maybe at least we also need it in other pallet's mock runtime for unit tests?

alannotnerd:
@kaichaosun What's your opinion? :p

primitives/src/network.rs
kaichaosun: The token decimals should never be used in runtime which lives on chain and hard to keep in sync with others chain.
For unit tests, hard code in test cases is fine to me.

alannotnerd: Got it. Fix it in next commit.

node/parallel/src/chain_spec/heiko.rs
kaichaosun: This is protocol-id right? if it's public network, should be `heiko`, public testnet `heiko-staging`, local network should be `heiko-dev` and `heiko-local`.

primitives/src/network.rs
yz89: @kaichaosun Every token has registered the decimal when applying the ss58 prefix. Can we use the token decimals as same as the declared in the substrate s58-registry.json? 
https://github.com/paritytech/substrate/pull/9254/files#diff-b501fa660c67a198b63b9d0ed792673bb94d91c275dfcf2d250ce06eafb66419R540

alannotnerd: I'm not sure, but a trivial solution is fetch and parse it in build script.

kaichaosun: I'm not keen to use the json file, as it's prepared for frontend if i remember correctly.

primitives/src/network.rs
```diff
@@ -0,0 +1,28 @@
...
+        }
+    }
+}
+
+pub const HEIKO_PREFIX: u8 = 110;
+pub const PARALLEL_PREFIX: u8 = 172;
+/// Token symbol of heiko network.
+pub const HEIKO_TOKEN: &str = "HKO";
+/// Token symbol of parallel network.
+pub const PARALLEL_TOKEN: &str = "PARA";
```
kaichaosun, line 28: This config better not put in primitives.

rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly
```
kaichaosun, line 1: It's better to keep consistent with Substrate which doesn't contains this file.

kaichaosun (approved):
Two more concerns, others are good to be merged.

primitives/src/network.rs
GopherJ: @alannotnerd you can use `stringify!(currencyId` to get it

yz89 (approved):

PR #323, opened 2021-07-23
simplify nominee election pallet

GopherJ:
companion pr: https://github.com/parallel-finance/nominate-client/pull/6 and https://github.com/parallel-finance/parallel-dapp/pull/342


runtime/heiko/src/lib.rs
```diff
@@ -958,11 +957,6 @@ type EnsureRootOrHalfCouncil = EnsureOneOf<
     EnsureRoot<AccountId>,
     pallet_collective::EnsureProportionMoreThan<_1, _2, AccountId, CouncilCollective>,
 >;
-type EnsureRootOrHalfTechnical = EnsureOneOf<
```
kaichaosun, line 961: Still need this origin for other pallets, like market params change.

kaichaosun (approved):
Well done. 👍

runtime/heiko/src/lib.rs
GopherJ: @kaichaosun yes, in last commit I'd like to use it for `ScheduleOrigin` but finally found that's in pallet-schedule it seems it hardcoded `ensure_root`, and it's the only place that I used EnsureRootOrHalfTechnical, so I remove it for now

GopherJ:
I removed whiltelist validator set because currently we don't have a clear plan for it

PR #336, opened 2021-07-30
update docs for runtime event - DON'T MERGE

yz89:
close https://github.com/parallel-finance/parallel-dapp/issues/298

Updated by https://www.notion.so/parallelfinance/Testnet-Error-Messages-FE-18b20df10d524027be0bffad39b1826a.

But I don't think it makes sense. The comments of the backend code will generate API docs like this https://api-docs.parallel.fi/pallet_loans/pallet/enum.Error.html.

@Helen8 Please reconsider these words from the perspective of the API documentation. Or the front end of the stand-alone speech.

### DIFF
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -124,33 +124,33 @@ pub mod pallet {
 
     #[pallet::error]
     pub enum Error<T> {
-        /// Insufficient liquidity to borrow more or disable collateral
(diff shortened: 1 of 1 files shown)


yz89:
@Helen8 We decided to separate the wording of the frontend and backend. Could you please add a new column in the doc to help us update the comments of the backend? https://www.notion.so/parallelfinance/Testnet-Error-Messages-FE-18b20df10d524027be0bffad39b1826a

PR #344, opened 2021-08-02
Use correct ss58 format in `frame_system::Config`

alannotnerd:
Respect the documentation of `frame_system::Config`. It said the associated type `SS58Format` would replace the properties in `chain_spec` which seems not😿.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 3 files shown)


GopherJ:
ICE is often caused by toolchain version

GopherJ:
@alannotnerd  maybe you can change CI to use an older version of rust-toolchain

alannotnerd:
Downgrade to 1.53, and it works for now

GopherJ (approved):

PR #346, opened 2021-08-02
Add weightinfo for builtin pallet

alannotnerd:
Replace `()` with `SubstrateWeight`

### DIFF
--- a/runtime/heiko/src/lib.rs
+++ b/runtime/heiko/src/lib.rs
@@ -230,7 +230,7 @@ impl frame_system::Config for Runtime {
     /// The data to be stored in an account.
     type AccountData = pallet_balances::AccountData<Balance>;
     /// Weight information for the extrinsics of this pallet.
-    type SystemWeightInfo = ();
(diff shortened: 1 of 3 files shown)


GopherJ (approved):

yz89:
We should benchmark the built-in module by using our own cloud hosting. @kaichaosun Is that right?

GopherJ:
> We should benchmark the built-in module by using our own cloud hosting. @kaichaosun Is that right?
yes, https://github.com/paritytech/statemint/blob/master/scripts/benchmarks.sh

runtime/heiko/src/lib.rs
```diff
@@ -253,7 +253,7 @@ impl orml_tokens::Config for Runtime {
     type Amount = Amount;
     type CurrencyId = CurrencyId;
     type OnDust = orml_tokens::TransferDust<Runtime, TreasuryAccount>;
-    type WeightInfo = ();
+    type WeightInfo = (); // Yes, it is the correct type binding.
```
kaichaosun, line 256: orml tokens has benchmark function, need to use the actual result.

Issue #340, opened 2021-08-01
implement basic insurance pool in liquidStaking


GopherJ:
this should cover maybe the part of xcm fees & slashes, if possible

PR #348, opened 2021-08-02
basic insurance pool

GopherJ:
the insurance pool should cover xcm fees and try to cover maximum slashes

close: #350 
close: #340 

it'll charge staking fees & xcm fees while users stake, then when the stake-client withdraw it'll try to pay the xcm fees, when slashes happen it'll also pay as much as possible.

when user unstake and in the final step, it'll also pay the xcm fees

**Important**: I'm not yet happy to use this approach as it makes the xcm amount more complex. I still suggest #355, I'll raise maybe new pr to solve #350, and use insurance pool to cover only slashes

the xcm fees we may need to find another solution

### DIFF
--- a/pallets/liquid-staking/src/lib.rs
+++ b/pallets/liquid-staking/src/lib.rs
@@ -624,6 +624,14 @@ pub mod pallet {
                 },
             )?;
 
+            let xcm_weight = T::BaseXcmWeight::get() as Balance;
+            TotalReserves::<T>::try_mutate(|b| -> DispatchResult {
+                *b = b
+                    .checked_sub(xcm_weight)
+                    .ok_or(ArithmeticError::Underflow)?;
+                Ok(())
+            })?;
+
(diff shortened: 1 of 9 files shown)


pallets/liquid-staking/src/lib.rs
```diff
@@ -261,11 +266,37 @@ pub mod pallet {
...
+                .mul_floor(
+                    amount
+                        .checked_sub(xcm_weight)
+                        .ok_or(ArithmeticError::Underflow)?,
+                )
+                .checked_add(xcm_weight)
+                .ok_or(ArithmeticError::Overflow)?;
+
+            TotalReserves::<T>::try_mutate(|b| -> DispatchResult {
+                *b = b.checked_add(reserves).ok_or(ArithmeticError::Overflow)?;
```
alannotnerd, line 288: Storage mutation should be invoked after all potienal exception checked.

Issue #350, opened 2021-08-03
dont burn xcm fees but deposit it to staking pool account


GopherJ:
https://github.com/paritytech/polkadot/blob/c4ee9d463adccfa3bf436433e3e26d0de5a4abbc/xcm/xcm-builder/src/weight.rs#L135-L140

PR #357, opened 2021-08-05
Feat/dont burn xcm fees

GopherJ:
close: #350

### DIFF
--- a/runtime/heiko/src/constants.rs
+++ b/runtime/heiko/src/constants.rs
@@ -74,6 +74,9 @@ pub mod fee {
             // in Kusama, extrinsic base weight (smallest non-zero weight) is mapped to 1/10 CENT:
             // in Statemine, it's mapped to 1/10 of that, or 1/100 CENT
             // in Heiko, it's mapped to 1/100 CENT
+            //
+            // coeff_integer * x^(degree) + coeff_frac * x^(degree)
+            //
             let p = super::currency::CENTS;
             let q = 100 * Balance::from(ExtrinsicBaseWeight::get());
             smallvec![WeightToFeeCoefficient {
--- a/runtime/heiko/src/constants.rs
+++ b/runtime/heiko/src/constants.rs
@@ -47,8 +47,8 @@ pub mod time {
 /// Fee-related.
 pub mod fee {
     use frame_support::weights::{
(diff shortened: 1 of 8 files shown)


runtime/heiko/src/constants.rs
```diff
@@ -72,7 +72,8 @@ pub mod fee {
         type Balance = Balance;
         fn polynomial() -> WeightToFeeCoefficients<Self::Balance> {
             // in Kusama, extrinsic base weight (smallest non-zero weight) is mapped to 1/10 CENT:
-            // in Statemine, we map to 1/10 of that, or 1/100 CENT
+            // in Statemine, it's mapped to 1/10 of that, or 1/100 CENT
```
kaichaosun, line 75: Can you elaborate on this `it's mapped to 1/10 of that, or 1/100 CENT`?

runtime/heiko/src/constants.rs
```diff
@@ -83,4 +84,11 @@ pub mod fee {
             }]
         }
     }
+
+    pub fn ksm_per_second() -> u128 {
```
kaichaosun, line 88: Can you share more on the fomula?

runtime/parallel/src/lib.rs
```diff
@@ -397,6 +398,21 @@ impl pallet_liquid_staking::Config for Runtime {
     type BaseXcmWeight = BaseXcmWeight;
 }
 
+parameter_types! {
+    pub StakingPoolAccount: AccountId = StakingPalletId::get().into_account();
+}
+
+pub struct ToStakingPool;
+impl TakeRevenue for ToStakingPool {
```
kaichaosun, line 406: Where does the revenue come from?

GopherJ: it comes from xcm dest weight @kaichaosun

runtime/heiko/src/constants.rs
GopherJ: @kaichaosun the formula will be:

```
coeff_integer * x^(degree) + coeff_frac * x^(degree)
```

`WeightToFeePolynomial` uses it to covert `weight` to balance

and here it's `1/100` CENT because:

```
let q = 100 * Balance::from(ExtrinsicBaseWeight::get());
```
q has been multiplied by 100

ayushmishra2005 (approved):

PR #358, opened 2021-08-05
use insurance pool to cover slashes

GopherJ:


pallets/liquid-staking/src/lib.rs
```diff
@@ -326,6 +329,11 @@ pub mod pallet {
                 T::BaseXcmWeight::get(),
             )?;
 
+            ensure!(
+                matches!(xcm_result, Outcome::Complete(_)),
+                Error::<T>::XcmTransferFailed
```
kaichaosun, line 334: If xcm transfer failed, does the fund returns back to the original account?

pallets/liquid-staking/src/lib.rs
```diff
@@ -390,16 +398,37 @@ pub mod pallet {
...
-            })?;
-            let exchange_rate = Rate::checked_from_rational(
-                TotalStakingAsset::<T>::get(),
-                TotalVoucher::<T>::get(),
-            )
-            .ok_or(Error::<T>::InvalidExchangeRate)?;
-            ExchangeRate::<T>::put(exchange_rate);
+            let reserves = Self::total_reserves();
+            let left_reserves = reserves.saturating_sub(amount);
+            let reduced_reserves = reserves
```
kaichaosun, line 403: This function about insurance pool is confused.
The changes in my mind is:
* create a pool, and deposits insurance fund from specific origin
* liquid-staking rely on an insurance provider (trait), when slash happens, use this insurance provider, the provider should be able to return the covered slashed funds, and the uncovered funds. If there are uncovered slash funds, then remove total staking asset and update exchange rate.
* (nice to have) the insurance funds can deposits in lending market and get interest.

pallets/liquid-staking/src/lib.rs
GopherJ: @kaichaosun it depends, if it fails on parachain, then yes, but if it fails on relaychain like:

```
FailedToTransactAsset 
```

then we lose them

pallets/liquid-staking/src/lib.rs
GopherJ: @kaichaosun the second point is actually covered already, we do calculate the covered slashes, but yes, we didn't allow any users to deposit funds.

PR #362, opened 2021-08-06
WIP: add LiquidStaking2.0

MrPai:
Develop LiquidStaking2.0 logic and test


pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+        /// ExchangeRate is invalid
+        InvalidExchangeRate,
+    }
+
+    #[pallet::event]
+    #[pallet::generate_deposit(pub(super) fn deposit_event)]
+    pub enum Event<T: Config> {
+        /// The assets get staked successfully
+        Staked(T::AccountId, Balance),
+        /// The voucher get unstaked successfully
```
ayushmishra2005, line 124: Either `The voucher gets unstaked successfully`

or `The vouchers get unstaked successfully`

yz89:
Regarding the pallet name, I prefer using liquid-staking-new for it might take a while.

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+        pub fn record_unbond_response(
+            origin: OriginFor<T>,
+            era_index: Option<EraIndex>,
+        ) -> DispatchResultWithPostInfo {
+            let _who = ensure_signed(origin)?;
+            let era_index = era_index.unwrap_or(Self::current_era());
+
+            let op = StakingOperationHistory::<T>::try_mutate(
+                &era_index,
+                StakingOperationType::Bond,
```
yz89, line 446: Shouldn't it be Unbond?

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+        })?;
+        Self::deposit_event(Event::Unstaked(who.clone(), amount, asset_amount));
+        Ok(().into())
+    }
+
+    fn claim(who: &T::AccountId) -> DispatchResultWithPostInfo {
+        let mut withdrawable_stake_amount = 0u128;
+        let mut withdrawable_unstake_amount = 0u128;
+        let mut remove = vec![];
+        MatchingQueue::<T>::iter_prefix(who).for_each(|(era_index, user_buffer)| {
```
yz89, line 649: It's such a long function, and there are too many if-else nesting. Replace code in the matching branch with a function and extract common check items.

yz89:
Need basic unit test for every function especially the extrinsics. @MrPai @alannotnerd

pallets/liquid-staking-two/src/lib.rs
alannotnerd: Sure, I'll fix it

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+            })?;
+
+            let exchange_rate = Rate::checked_from_rational(
+                StakingPool::<T>::get(),
+                T::Currency::total_issuance(T::LiquidCurrency::get()),
+            )
+            .ok_or(Error::<T>::InvalidExchangeRate)?;
+            ExchangeRate::<T>::put(exchange_rate);
+
+            StakingOperationHistory::<T>::insert(
```
GopherJ, line 347: the `StakingOperationHistory` struct doesn't make too much sense here, in one era we can have multiple rewards from multiple agents

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
```
GopherJ, line 388: same as above

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+        ) -> DispatchResultWithPostInfo {
+            let _who = ensure_signed(origin)?;
+            let era_index = era_index.unwrap_or(Self::current_era());
+
+            let op = StakingOperationHistory::<T>::try_mutate(
+                era_index,
+                StakingOperationType::Bond,
+                |op| -> Result<Operation<_>, DispatchError> {
+                    let next_op = op
+                        .clone()
```
GopherJ, line 417: this clone can be avoid

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+                        .filter(|op| op.status == ResponseStatus::Ready)
+                        .map(|op| Operation {
+                            status: ResponseStatus::Successed,
+                            ..op
+                        })
+                        .ok_or(Error::<T>::OperationNotReady)?;
+                    *op = Some(next_op.clone());
```
GopherJ, line 424: if we only need `amout`, we should remove this `clone` and return back only `amount`

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+        ) -> DispatchResultWithPostInfo {
+            let _who = ensure_signed(origin)?;
+            let era_index = era_index.unwrap_or(Self::current_era());
+
+            let op = StakingOperationHistory::<T>::try_mutate(
+                &era_index,
+                StakingOperationType::Unbond,
+                |op| -> Result<Operation<_>, DispatchError> {
+                    let next_op = op
+                        .clone()
```
GopherJ, line 449: we should remove this clone

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+                        .filter(|op| op.status == ResponseStatus::Ready)
+                        .map(|op| Operation {
+                            status: ResponseStatus::Successed,
+                            ..op
+                        })
+                        .ok_or(Error::<T>::OperationNotReady)?;
+                    *op = Some(next_op.clone());
+                    Ok(next_op)
```
GopherJ, line 457: pls return back `amount` instead

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+            Ok(().into())
+        }
+
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn transfer_to_relaychain(
+            origin: OriginFor<T>,
+            amount: Balance,
+            agent: T::AccountId,
+        ) -> DispatchResultWithPostInfo {
```
GopherJ, line 471: check agent is in `Members`

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+            Ok(().into())
+        }
+
+        /// Invoked when unbonding extrinsic finished. Burn previously transfered xksm if
+        /// successed.
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn record_unbond_response(
+            origin: OriginFor<T>,
+            era_index: Option<EraIndex>,
```
GopherJ, line 439: `EraIndex` needs to be ensured, stake-client will not feed option value because it can cause errors

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+                    pool_unstake_amount - pool_stake_amount,
+                )
+            } else {
+                (StakingOperationType::Matching, 0)
+            };
+            StakingOperationHistory::<T>::try_mutate(
+                &previous_era,
+                &operation_type,
+                |o| -> DispatchResult {
+                    ensure!(*o == None, "error");
```
GopherJ, line 518: pls define an error

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+            #[pallet::compact] amount: Balance,
+        ) -> DispatchResultWithPostInfo {
+            let who = ensure_signed(origin)?;
+            <Self as LiquidStakingProtocol<T::AccountId>>::unstake(&who, amount)?;
+            Ok(().into())
+        }
+
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn claim(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
```
GopherJ, line 567: DispatchResult should be preferred if we don't return anything

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+        }
+
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn unstake(
+            origin: OriginFor<T>,
+            #[pallet::compact] amount: Balance,
+        ) -> DispatchResultWithPostInfo {
+            let who = ensure_signed(origin)?;
+            <Self as LiquidStakingProtocol<T::AccountId>>::unstake(&who, amount)?;
```
GopherJ, line 561: DispatchResult should be preferred if we don't return anything

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+            ));
+            Ok(().into())
+        }
+
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn stake(
+            origin: OriginFor<T>,
+            #[pallet::compact] amount: Balance,
+        ) -> DispatchResultWithPostInfo {
```
GopherJ, line 548: DispatchResult should be preferred if we don't return anything

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+        MatchingUserBuffer,
+        ValueQuery,
+    >;
+
+    /// Previous era index on Parachain.
+    ///
+    /// Considering that in case stake client cannot update era index one by one.
+    #[pallet::storage]
+    #[pallet::getter(fn previous_era)]
+    pub type PreviousEra<T: Config> = StorageValue<_, EraIndex, ValueQuery>;
```
GopherJ, line 251: ValueQuery will give by default era 0, let's check if relaychain starts from `0` or `1`

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T>
+    where
+        [u8; 32]: From<<T as frame_system::Config>::AccountId>,
+    {
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn trigger_new_era(
```
GopherJ, line 309: we should use this function only to update `CurrentEra` or `PreviousEra`, another function is needed to handle matching result and do real operations

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+                    return None;
+                }
+                let current_era = Self::current_era();
+                match t {
+                    StakingOperationType::Bond => {
+                        // if bond, need to wait 1 era or get the matching part instantly
+                        if era_index + T::StakingDuration::get() > current_era {
+                            if !user_buffer.claimed_matching {
+                                // get the matching part only
+                                MatchingQueue::<T>::try_mutate(
```
GopherJ, line 665: if no error returned, pls use `mutate`

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+
+    fn claim(who: &T::AccountId) -> DispatchResultWithPostInfo {
+        let mut withdrawable_stake_amount = 0u128;
+        let mut withdrawable_unstake_amount = 0u128;
+        let mut remove = vec![];
+        MatchingQueue::<T>::iter_prefix(who).for_each(|(era_index, user_buffer)| {
+            let mut claim_unstake_each_era = 0u128;
+            let mut claim_stake_each_era = 0u128;
+            let pool_buffer = MatchingPool::<T>::get(&era_index);
+            pool_buffer.operation_type.and_then(|t| {
```
GopherJ, line 653: it doesn't make sense to put a big function into `and_then`

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+#[derive(Copy, Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+pub struct Operation<BlockNumber> {
+    amount: Balance,
+    block_number: BlockNumber,
+    status: ResponseStatus,
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Default, Encode, Decode, RuntimeDebug)]
+pub struct MatchingPoolBuffer {
+    total_unstake_amount: Balance,
```
GopherJ, line 79: let's use compacted Balance

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+pub enum ResponseStatus {
+    Ready,
+    Processing,
+    Successed,
+    Failed,
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+pub struct Operation<BlockNumber> {
+    amount: Balance,
```
GopherJ, line 72: compacted balance should be preferred

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+    total_stake_amount: Balance,
+    operation_type: Option<StakingOperationType>,
+}
+
+/// The single user's stake/unsatke amount in each era
+#[derive(Copy, Clone, Eq, PartialEq, Default, Encode, Decode, RuntimeDebug)]
+pub struct MatchingUserBuffer {
+    /// The token amount that user unstake during this era, will be calculated
+    /// by exchangerate and xToken amount
+    total_unstake_amount: Balance,
```
GopherJ, line 89: compacted balance should be preferred

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+
+#[derive(Copy, Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+pub struct Operation<BlockNumber> {
+    amount: Balance,
+    block_number: BlockNumber,
+    status: ResponseStatus,
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Default, Encode, Decode, RuntimeDebug)]
+pub struct MatchingPoolBuffer {
```
GopherJ, line 78: Pool and Buffer have similar meaning

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+    total_unstake_amount: Balance,
+    total_stake_amount: Balance,
+    operation_type: Option<StakingOperationType>,
+}
+
+/// The single user's stake/unsatke amount in each era
+#[derive(Copy, Clone, Eq, PartialEq, Default, Encode, Decode, RuntimeDebug)]
+pub struct MatchingUserBuffer {
```
GopherJ, line 86: Pool and Buffer have similar meaning

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+            let _who = ensure_signed(origin)?;
+            // ensure!(T::Members::contains(&agent), Error::<T>::IllegalAgent);
+            StakingPool::<T>::try_mutate(|m| -> DispatchResult {
+                *m = m.checked_add(amount).ok_or(ArithmeticError::Overflow)?;
+                Ok(())
+            })?;
+
+            let exchange_rate = Rate::checked_from_rational(
+                StakingPool::<T>::get(),
+                T::Currency::total_issuance(T::LiquidCurrency::get()),
```
GopherJ, line 342: so we dont want to record `TotalVoucher` but will use xKSM's balance? what's the reason of this one?

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+        pub fn record_reward(
+            origin: OriginFor<T>,
+            agent: T::AccountId,
+            era_index: EraIndex,
+            #[pallet::compact] amount: Balance,
+        ) -> DispatchResultWithPostInfo {
+            // T::WithdrawOrigin::ensure_origin(origin)?;
+            let _who = ensure_signed(origin)?;
+            // ensure!(T::Members::contains(&agent), Error::<T>::IllegalAgent);
+            StakingPool::<T>::try_mutate(|m| -> DispatchResult {
```
GopherJ, line 335: we should check if this agent has already one successful reward in this era, if yes then refuse the new one

GopherJ: maybe the staking pool mutation should be done after this check

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+        /// Can only invoked by stake client. Decrease asset amount in `StakingPool` and change
+        /// exchange rate.
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn record_slash(
+            origin: OriginFor<T>,
+            agent: T::AccountId,
+            era_index: EraIndex,
+            #[pallet::compact] amount: Balance,
+        ) -> DispatchResultWithPostInfo {
```
GopherJ, line 372: let's limit the maxmium slash of one agent per N block

GopherJ: Question: will we have multiple slashes for the same agent in the same era?

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+                    Ok(next_op)
+                },
+            )?;
+            T::Currency::withdraw(T::LiquidCurrency::get(), &Self::account_id(), op.amount)?;
+            Self::deposit_event(Event::<T>::UnbondSucceed(era_index));
+            Ok(().into())
+        }
+
+        #[pallet::weight(10_000)]
+        #[transactional]
```
GopherJ, line 466: this function alone is useless now because we can already call xTokens to do this

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+
+impl<T: Config> LiquidStakingProtocol<T::AccountId> for Pallet<T> {
+    // After confirmed bond on relaychain,
+    // after update exchangerate (record_reward),
+    // and then mint/deposit xKSM.
+    fn stake(who: &T::AccountId, amount: Balance) -> DispatchResultWithPostInfo {
+        //todo reserve, insurance pool
+        T::Currency::transfer(T::StakingCurrency::get(), who, &Self::account_id(), amount)?;
+
+        MatchingQueue::<T>::try_mutate(
```
GopherJ, line 583: I'm a bit lost in the naming `matchingpool` & `matchingqueue`

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+                .checked_add(asset_amount)
+                .ok_or(ArithmeticError::Overflow)?;
+
+            Ok(())
+        })?;
+        Self::deposit_event(Event::Unstaked(who.clone(), amount, asset_amount));
+        Ok(().into())
+    }
+
+    fn claim(who: &T::AccountId) -> DispatchResultWithPostInfo {
```
GopherJ, line 645: it's not good that one claim will cause us to iterate the whole queue

GopherJ:
I didn't review `claim` function yet because it's too long and should be refactored, I'll review again this part after the refactoring

pallets/liquid-staking-two/src/lib.rs
MrPai: On-chain will not store rewards for each agent, besides on relaychain, reward will be paid by era.
so if we have multiple agents on relaychain, let's calculate their reward into one value, then push to parachain.

GopherJ:
staging files are now located at: https://github.com/parallel-finance/static-files

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+            Self::deposit_event(Event::<T>::SlashRecorded(agent, amount));
+            Ok(().into())
+        }
+
+        // bond/unbond/rebond/bond_extra may be merge into one
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn record_bond_response(
+            origin: OriginFor<T>,
+            era_index: Option<EraIndex>,
```
GopherJ, line 407: this shouldn't be Option, stake-client will ensure to feed era index

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+                T::BaseXcmWeight::get(),
+            )?;
+            Ok(().into())
+        }
+
+        // this method must be called after update the era index, so we use previous era to calculate
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn deposit_event_to_relaychain(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
+            //todo ensure that this method is called after update era index.
```
GopherJ, line 494: `todo` will not be recognized by editor/ide, we should use `TODO` instead

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+                ),
+                T::BaseXcmWeight::get(),
+            )?;
+            Ok(().into())
+        }
+
+        // this method must be called after update the era index, so we use previous era to calculate
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn deposit_event_to_relaychain(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
```
GopherJ, line 493: shall this function be called before next era and use current era? so that users can start to earn rewards in next era

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+                            .saturating_sub(user_buffer.claimed_unstake_amount);
+
+                        claim_stake_each_era = user_buffer
+                            .total_stake_amount
+                            .saturating_sub(user_buffer.claimed_stake_amount);
+                    }
+                    _ => (),
+                };
+
+                MatchingQueue::<T>::try_mutate(who, &era_index, |b| -> DispatchResult {
```
GopherJ, line 749: if there won't be error thrown, we should prefer `mutate`

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+
+use orml_traits::{MultiCurrency, MultiCurrencyExtended};
+
+pub use pallet::*;
+use primitives::{
+    Amount, Balance, CurrencyId, EraIndex, ExchangeRateProvider,
+    LiquidStakingProtocol, Rate, Ratio,
+};
+
+pub const MAX_UNSTAKE_CHUNKS: usize = 5;
```
GopherJ, line 49: can you add a comment to explain what is this for?

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+
+        /// Approved agent list on relaychain
+        // type Members: SortedMembers<Self::AccountId>;
+
+        /// Base xcm weight to use for cross chain transfer
+        type BaseXcmWeight: Get<Weight>;
+
+        /// The maximum size of Unstake
+        // #[pallet::constant]
+        type MaxUnstake: Get<u32>;
```
GopherJ, line 162: where will this be used?

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+    Unbond,
+    Rebond,
+    Matching,
+    TransferToRelaychain,
+    RecordReward,
+    RecordSlash,
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+pub enum ResponseStatus {
```
GopherJ, line 63: can you give some comments on the meaning of every status and how stake-client should use it?

pallets/liquid-staking-two/src/lib.rs
GopherJ: but in the parameters, we have `agent`

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+        #[pallet::constant]
+        type PalletId: Get<PalletId>;
+
+        /// Number of eras that user can get xToken back after stake on parachain.
+        #[pallet::constant]
+        type StakingDuration: Get<EraIndex>;
+
+        /// Number of eras that staked funds must remain bonded for.
+        #[pallet::constant]
+        type BondingDuration: Get<EraIndex>;
```
GopherJ, line 146: this is a constant which means if relaychain changed this value, we also need to do runtime upgrade?

primitives/src/lib.rs
```diff
@@ -149,3 +152,10 @@ pub trait EmergencyPriceFeeder<CurrencyId, PriceWithDecimal> {
 pub trait ExchangeRateProvider {
     fn get_exchange_rate() -> Rate;
 }
+
+//todo change the return type
+pub trait LiquidStakingProtocol<AccountId> {
```
kaichaosun, line 157: Extract this trait doesn't make sense. It's just user's operation.

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+    }
+}
+
+impl<T: Config> Pallet<T> {
+    pub fn account_id() -> T::AccountId {
+        T::PalletId::get().into_account()
+    }
+}
+
+impl<T: Config> ExchangeRateProvider for Pallet<T> {
```
kaichaosun, line 816: This can be easily added if need.

kaichaosun (commented):
General feedback for the new design and implementation
* if still want use agent (multi-sig wallets) to manage staking workflow, it's better to add new functionalities piece by piece in the current liquid-staking-pallet.
* if want to use `xcm::transact` to make a trustless workflow, better to make such messages test passed.

I maybe wrong here as I'm not familiar with the new business requirements.

pallets/liquid-staking-two/src/lib.rs
alannotnerd: `Members` has been removed.

pallets/liquid-staking-two/src/lib.rs
alannotnerd: Respect era_index passed by stake-client.

pallets/liquid-staking-two/src/lib.rs
alannotnerd: Yes, we have another function.

pallets/liquid-staking-two/src/lib.rs
alannotnerd: @MrPai

pallets/liquid-staking-two/src/lib.rs
alannotnerd: Yeah, it mutates only if the status is `Ready`.

pallets/liquid-staking-two/src/lib.rs
alannotnerd: I guess compacted balance should only be used in extrinsic signature.

pallets/liquid-staking-two/src/lib.rs
alannotnerd: Removed

pallets/liquid-staking-two/src/lib.rs
alannotnerd: Perferred `DispatchResultWithPostInfo` within an extrinsic call. Use `DispatchResult` in other funtion.

pallets/liquid-staking-two/src/lib.rs
alannotnerd: Same upon.

pallets/liquid-staking-two/src/lib.rs
alannotnerd: Same upon

pallets/liquid-staking-two/src/lib.rs
alannotnerd: @MrPai

pallets/liquid-staking-two/src/lib.rs
alannotnerd: Currently use associated type instead

pallets/liquid-staking-two/src/lib.rs
alannotnerd: same upon

primitives/src/lib.rs
alannotnerd: @MrPai

pallets/liquid-staking-two/src/types.rs
```diff
@@ -0,0 +1,110 @@
...
+                StakingOperationType::Bond,
+                self.total_stake_amount - self.total_unstake_amount,
+            )
+        } else if self.total_stake_amount < self.total_unstake_amount {
+            (
+                StakingOperationType::Unbond,
+                self.total_unstake_amount - self.total_stake_amount,
+            )
+        } else {
+            (StakingOperationType::Matching, 0)
```
GopherJ, line 53: I think `Matching` can be renamed to `NoOp` etc

pallets/liquid-staking-two/src/types.rs
```diff
@@ -0,0 +1,110 @@
...
+//TODO(wangyafei): use associated type.
+use primitives::{Balance, Rate};
+
+#[derive(Copy, Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+pub enum StakingOperationType {
+    Bond,
+    Unbond,
+    Rebond,
+    Matching,
+    TransferToRelaychain,
```
GopherJ, line 13: we dont need `TransferToRelayChain` because xTokens already offer it

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+    where
+        [u8; 32]: From<<T as frame_system::Config>::AccountId>,
+    {
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn trigger_new_era(
+            origin: OriginFor<T>,
+            era_index: EraIndex,
+        ) -> DispatchResultWithPostInfo {
+            let _who = ensure_signed(origin)?;
```
GopherJ, line 235: we should add a TODO here to limit who can trigger_new_era

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+            Self::deposit_event(Event::<T>::EraUpdated(era_index));
+            Ok(().into())
+        }
+
+        /// Record reward on each era, invoked by stake-client
+        /// StakingPool = StakingPool + reward amount
+        /// StakingPool/T::currency::total_issuance(liquidcurrency)
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn record_reward(
```
GopherJ, line 248: I think maybe it should be renamed to `record_rewards`

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+                    block_number: frame_system::Pallet::<T>::block_number(),
+                    status: ResponseStatus::Successed,
+                },
+            );
+
+            Self::deposit_event(Event::<T>::SlashRecorded(amount));
+            Ok(().into())
+        }
+
+        /// Invoked when unbonding extrinsic finished. Mint xksm if
```
GopherJ, line 303: unbonding -> bond

pallets/liquid-staking-two/src/lib.rs
MrPai: in relaychain, the default value is 0;
https://github.com/paritytech/substrate/blob/master/frame/staking/src/pallet/mod.rs#L712

pallets/liquid-staking-two/src/lib.rs
MrPai: `T::Currency::total_issuance(T::LiquidCurrency::get()), ` is same with TotalVoucher

pallets/liquid-staking-two/src/lib.rs
MrPai: we should transfer assets from liquidStaking pallet, so the method is necessary

MrPai: we should transfer assets from liquidStaking pallet, so the method is necessary

pallets/liquid-staking-two/src/lib.rs
MrPai: it claimed according to accountId, not the whole queue.

pallets/liquid-staking-two/src/lib.rs
MrPai: good point

pallets/liquid-staking-two/src/lib.rs
MrPai: it is used for calculate the previous era.

pallets/liquid-staking-two/src/lib.rs
MrPai: I think when relaychain do runtime upgrade, parachain should do runtime upgrade too.

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+    // considering users who stake and forget to claim within 28 eras.
+    fn claim_in_unbond_operation(
+        who: &T::AccountId,
+        claim_era: &EraIndex,
+        current_era: &EraIndex,
+        user_ledger_per_era: &UserLedgerPerEra,
+        pool_ledger_per_era: &PoolLedgerPerEra,
+    ) -> WithdrawalAmount {
+        if claim_era.clone() + T::BondingDuration::get() > current_era.clone() {
+            if !user_ledger_per_era.claimed_matching {
```
GopherJ, line 535: we should avoid nested if, quick return should be preferred

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+    // after waiting 1 era, users who stake get the left part
+    fn claim_in_bond_operation(
+        who: &T::AccountId,
+        claim_era: &EraIndex,
+        current_era: &EraIndex,
+        user_ledger_per_era: &UserLedgerPerEra,
+        pool_ledger_per_era: &PoolLedgerPerEra,
+    ) -> WithdrawalAmount {
+        if claim_era.clone() + T::StakingDuration::get() > current_era.clone() {
+            if !user_ledger_per_era.claimed_matching {
```
GopherJ, line 510: quick return should be preferred

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+            }
+        } else {
+            user_ledger_per_era.remaining_withdrawal_limit()
+        }
+    }
+}
+
+impl<T: Config> Pallet<T> {
+    #[inline]
+    fn increase_staked_asset(amount: BalanceOf<T>) -> DispatchResult {
```
GopherJ, line 552: we can put update_exchange_rate directly inside this function

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+    fn try_mark_op_succeed(
+        era_index: EraIndex,
+        op_type: StakingOperationType,
+    ) -> Result<BalanceOf<T>, DispatchError> {
+        StakingOperationHistory::<T>::try_mutate(
+            era_index,
+            op_type,
+            |op| -> Result<BalanceOf<T>, DispatchError> {
+                let next_op = op
+                    .filter(|op| op.status == ResponseStatus::Ready)
```
GopherJ, line 588: seems this will remove all other operations `notReady`

pallets/liquid-staking-two/src/protocol.rs
```diff
@@ -0,0 +1,140 @@
...
+                era_index,
+                user_ledger_per_era,
+                &mut withdrawable_stake_amount,
+                &mut withdrawable_unstake_amount,
+                &mut remove_record_from_user_queue,
+            );
+        });
+
+        // remove finished records from MatchingQueue
+        if remove_record_from_user_queue.len() > 0 {
```
GopherJ, line 109: `!is_empty` should be preferred

pallets/liquid-staking-two/src/protocol.rs
```diff
@@ -0,0 +1,140 @@
...
+            remove_record_from_user_queue.iter().for_each(|era_index| {
+                MatchingQueueByUser::<T>::remove(who, era_index);
+            });
+        }
+
+        // transfer xKSM from palletId to who
+        if withdrawable_stake_amount > 0 {
```
GopherJ, line 116: `!is_zero` should be preferred

pallets/liquid-staking-two/src/protocol.rs
```diff
@@ -0,0 +1,140 @@
...
+            T::Currency::transfer(
+                T::LiquidCurrency::get(),
+                &Self::account_id(),
+                who,
+                xtoken_amount,
+            )?;
+        }
+
+        // transfer KSM from palletId to who
+        if withdrawable_unstake_amount > 0 {
```
GopherJ, line 130: `!is_zero` should be preferred

GopherJ:
@MrPai  @alannotnerd  can we rename it to be: `liquid-staking-v2` and add it to runtime?

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+        user_ledger_per_era: UserLedgerPerEra,
+        withdrawable_unstake_amount: &mut u128,
+        withdrawable_stake_amount: &mut u128,
+        remove_record_from_user_queue: &mut Vec<EraIndex>,
+    ) {
+        let pool_ledger_per_era = MatchingPoolByEra::<T>::get(&era_index);
+        let success_operation = pool_ledger_per_era.operation_type.and_then(|t| {
+            let operation = StakingOperationHistory::<T>::get(&era_index, &t)?;
+            if operation.status == ResponseStatus::Successed {
+                return Some(t);
```
GopherJ, line 450:
```rust
        let success_operation = pool_ledger_per_era.operation_type.and_then(|t| {
            StakingOperationHistory::<T>::get(&era_index, &t)
                .filter(|op| op.status == ResponseStatus::Successed)
        });
```

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+            }
+            None
+        });
+
+        if let Some(staking_operation_type) = success_operation {
```
GopherJ, line 455: pls quick return and avoid this sub scope

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+        .ok_or(Error::<T>::InvalidExchangeRate)?;
+        ExchangeRate::<T>::put(exchange_rate);
+        Ok(())
+    }
+
+    #[inline]
+    fn try_mark_op_succeed(
+        era_index: EraIndex,
+        op_type: StakingOperationType,
+    ) -> Result<BalanceOf<T>, DispatchError> {
```
GopherJ, line 582: this is `DispatchResult`

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+    {
+        #[pallet::weight(10_000)]
+        #[transactional]
+        pub fn trigger_new_era(
+            origin: OriginFor<T>,
+            era_index: EraIndex,
+        ) -> DispatchResultWithPostInfo {
+            let _who = ensure_signed(origin)?;
+            let current_era = Self::current_era();
+            PreviousEra::<T>::put(current_era);
```
GopherJ, line 237: we should also check if:

```
era_index > current_era
```

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+    // so all user who unstake in this era, no need wait 28 eras.
+    // after waiting 1 era, users who stake get the left part
+    fn claim_in_bond_operation(
+        who: &T::AccountId,
+        claim_era: &EraIndex,
+        current_era: &EraIndex,
+        user_ledger_per_era: &UserLedgerPerEra,
+        pool_ledger_per_era: &PoolLedgerPerEra,
+    ) -> WithdrawalAmount {
+        if claim_era.clone() + T::StakingDuration::get() > current_era.clone() {
```
GopherJ, line 509: can you try to see if:

```rust
claim_era + &T::StakingDuration::get() > current_era
```

works so that we remove two clone

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+    // after waiting 28 eras, users who unstake and get the left part
+    // considering users who stake and forget to claim within 28 eras.
+    fn claim_in_unbond_operation(
+        who: &T::AccountId,
+        claim_era: &EraIndex,
+        current_era: &EraIndex,
+        user_ledger_per_era: &UserLedgerPerEra,
+        pool_ledger_per_era: &PoolLedgerPerEra,
+    ) -> WithdrawalAmount {
+        if claim_era.clone() + T::BondingDuration::get() > current_era.clone() {
```
GopherJ, line 534: even if the previous advice doesn't work , you can still remove these two clone, because `u32` implemented `Copy`

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+
+    #[inline]
+    fn try_mark_op_succeed(
+        era_index: EraIndex,
+        op_type: StakingOperationType,
+    ) -> Result<BalanceOf<T>, DispatchError> {
+        StakingOperationHistory::<T>::try_mutate(
+            era_index,
+            op_type,
+            |op| -> Result<BalanceOf<T>, DispatchError> {
```
GopherJ, line 586: `DispatchResult` should be preferred

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+        era_index: EraIndex,
+        user_ledger_per_era: UserLedgerPerEra,
+        withdrawable_unstake_amount: &mut u128,
+        withdrawable_stake_amount: &mut u128,
+        remove_record_from_user_queue: &mut Vec<EraIndex>,
+    ) {
+        let pool_ledger_per_era = MatchingPoolByEra::<T>::get(&era_index);
+        let success_operation = pool_ledger_per_era.operation_type.and_then(|t| {
+            let operation = StakingOperationHistory::<T>::get(&era_index, &t)?;
+            if operation.status == ResponseStatus::Successed {
```
GopherJ, line 449:
```rust
        let success_operation = pool_ledger_per_era.operation_type.and_then(|t| {
            StakingOperationHistory::<T>::get(&era_index, &t)
                .filter(|op| op.status == ResponseStatus::Successed)
        });
```

pallets/liquid-staking-two/src/protocol.rs
```diff
@@ -0,0 +1,140 @@
...
+            },
+        )?;
+        Ok(asset_amount)
+    }
+
+    fn claim(who: &T::AccountId) -> DispatchResult {
+        let mut withdrawable_stake_amount = 0u128;
+        let mut withdrawable_unstake_amount = 0u128;
+        let mut remove_record_from_user_queue = Vec::<EraIndex>::new();
+        MatchingQueueByUser::<T>::iter_prefix(who).for_each(|(era_index, user_ledger_per_era)| {
```
GopherJ, line 97: we can use `fold` here, no need to declare https://github.com/parallel-finance/parallel/pull/362/files#diff-b81993106286609911b130de00ce1f6c1f3b5d7e7f6a71afbfd87989dfde742dR94-R96 and pass to accumulate_claim_by_era

primitives/src/lib.rs
MrPai: the trait is for the stability pool

MrPai:
> General feedback for the new design and implementation
> 
> * if still want use agent (multi-sig wallets) to manage staking workflow, it's better to add new functionalities piece by piece in the current liquid-staking-pallet.
> * if want to use `xcm::transact` to make a trustless workflow, better to make such messages test passed.
> 
> I maybe wrong here as I'm not familiar with the new business requirements.

there will be a stability pool and matching pool, and also we need users to claim their asset, not transfer by stake-client, so I think it may be difficult to add it piece by piece? but if you have any detailed info, solutions, or paths, please let us know.

pallets/liquid-staking-two/src/types.rs
MrPai: KSM can only be transfered from this pallet, not xTokens pallet,

pallets/liquid-staking-two/src/protocol.rs
MrPai: what's the different?

pallets/liquid-staking-two/src/lib.rs
MrPai: what's the different between `DispatchResult` and `Result`?

GopherJ (changes requested):

pallets/liquid-staking-two/src/lib.rs
GopherJ: no sorry DispatchResult's return value is `()` we cannot have `DispatchResult<T>`

pallets/liquid-staking-two/src/protocol.rs
GopherJ: we don't have hard coded number in the code, you can check loans pallet

pallets/liquid-staking-two/src/lib.rs
MrPai: we absolutely need it, why remove it?
https://github.com/parallel-finance/parallel/blob/master/runtime/heiko/src/lib.rs#L805

pallets/liquid-staking-two/src/protocol.rs
MrPai: https://doc.rust-lang.org/stable/src/alloc/vec/mod.rs.html#1842

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+        /// The withdraw request is successful
+        Claimed(T::AccountId),
+        /// The rewards are recorded
+        RewardsRecorded(BalanceOf<T>),
+        /// The slash is recorded
+        SlashRecorded(BalanceOf<T>),
+
+        DepositEventToRelaychain(T::AccountId, EraIndex, StakingOperationType, BalanceOf<T>),
+        /// Bond operation in relaychain was successed.
+        BondSucceed(EraIndex),
```
GopherJ, line 138: `BondSucceeded`

pallets/liquid-staking-two/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+        /// Unbond operation in relaychain was successed.
+        UnbondSucceed(EraIndex),
```
GopherJ, line 140: `UnbondSucceeded`

pallets/liquid-staking-two/src/lib.rs
GopherJ: you are right

pallets/liquid-staking-two/src/lib.rs
GopherJ, line 509: `EraIndex` should be passed because `EraIndex` implemented Copy and it's cheap, we don't need to clone manually

pallets/liquid-staking-two/src/protocol.rs
```diff
@@ -0,0 +1,140 @@
...
+    fn stake(who: &AccountId, amount: Balance) -> DispatchResult;
+    fn unstake(who: &AccountId, amount: Balance) -> Result<Balance, DispatchError>;
+    fn claim(who: &AccountId) -> DispatchResult;
+}
+
+impl<T: Config> LiquidStakingProtocol<T::AccountId, BalanceOf<T>> for Pallet<T> {
+    // After confirmed bond on relaychain,
+    // after update exchangerate (record_reward),
+    // and then mint/deposit xKSM.
+    fn stake(who: &T::AccountId, amount: BalanceOf<T>) -> DispatchResult {
```
GopherJ, line 23: https://substrate.dev/rustdocs/latest/frame_support/attr.require_transactional.html

let's add `#[require_transactional]`

pallets/liquid-staking-two/src/protocol.rs
```diff
@@ -0,0 +1,140 @@
...
+                Ok(())
+            },
+        )?;
+        Ok(().into())
+    }
+
+    // After confirmed unbond on relaychain,
+    // and then burn/withdraw xKSM.
+    // before update exchangerate (record_reward)
+    fn unstake(who: &T::AccountId, amount: BalanceOf<T>) -> Result<BalanceOf<T>, DispatchError> {
```
GopherJ, line 57: https://substrate.dev/rustdocs/latest/frame_support/attr.require_transactional.html

let's add `#[require_transactional]`

pallets/liquid-staking-two/src/protocol.rs
```diff
@@ -0,0 +1,140 @@
...
+                    .checked_add(asset_amount)
+                    .ok_or(ArithmeticError::Overflow)?;
+
+                Ok(())
+            },
+        )?;
+        Ok(asset_amount)
+    }
+
+    fn claim(who: &T::AccountId) -> DispatchResult {
```
GopherJ, line 93: https://substrate.dev/rustdocs/latest/frame_support/attr.require_transactional.html

let's add `#[require_transactional]`

pallets/liquid-staking-two/src/protocol.rs
GopherJ: yes but `!is_empty` is easier to read

pallets/liquid-staking-two/src/lib.rs
alannotnerd: I think it's better to compose them.

pallets/liquid-staking-two/src/lib.rs
alannotnerd: u r right. I'll do it in next commit.

pallets/liquid-staking-two/src/protocol.rs
```diff
@@ -0,0 +1,140 @@
...
+
+use crate::{Config, Error, ExchangeRate, MatchingPoolByEra, MatchingQueueByUser, Pallet};
+
+pub(crate) type BalanceOf<T> =
+    <<T as Config>::Currency as MultiCurrency<<T as frame_system::Config>::AccountId>>::Balance;
+
+//todo change the return type
+pub trait LiquidStakingProtocol<AccountId, Balance> {
+    fn stake(who: &AccountId, amount: Balance) -> DispatchResult;
+    fn unstake(who: &AccountId, amount: Balance) -> Result<Balance, DispatchError>;
```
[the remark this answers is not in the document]
alannotnerd: `DispatchResult` is alias of `Result<(), DispatchError>`. For this case, we need it to return a `Balance`.

pallets/liquid-staking-two/src/types.rs
```diff
@@ -0,0 +1,110 @@
...
+    RecordReward,
+    RecordSlash,
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+pub enum ResponseStatus {
+    Ready,
```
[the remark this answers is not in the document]
alannotnerd: Maybe could change `Ready` to `Pending`.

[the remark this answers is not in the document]
GopherJ: I also thought about `Pending` but maybe there are other better options, if no then we can use it

pallets/liquid-staking-two/src/lib.rs
GopherJ: you can also call directly `update_exchange_rate` in these two functions, no need to do it seperately as `reward` and `slash` will update exchange rate

pallets/liquid-staking-two/src/protocol.rs
alannotnerd: I agree `is_empty` is better choice in upon conversation, but not for there. The `Zero` and `One` are designed to set/clear flag but not to compare. I think it's okay here.

pallets/liquid-staking-two/src/lib.rs
alannotnerd: Much complicated. They're declared as reference in parameters. I've changed them passed by value.

pallets/liquid-staking-two/src/protocol.rs
alannotnerd: Sounds great! But I'm not sure the vector will be cloned or just moved during each step.

pallets/liquid-staking-v2/src/types.rs
```diff
@@ -0,0 +1,131 @@
...
+
+impl<Balance> PoolLedger<Balance>
+where
+    Balance: AtLeast32BitUnsigned + FullCodec + Copy + MaybeSerializeDeserialize + Debug + Default,
+{
+    pub fn op_after_new_era(&self) -> (StakingOperationType, Balance) {
+        if self.total_stake_amount > self.total_unstake_amount {
+            (
+                StakingOperationType::Bond,
+                self.total_stake_amount - self.total_unstake_amount,
```
GopherJ, line 51: these two currencies cannot be calculated like this, exchange rate must be used

pallets/liquid-staking-v2/src/types.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    // after matching mechanism，for bond operation, user who unstake can get all amount directly
+    // and user who stake only get the matching part
+    pub fn instant_withdrawal_by_bond(
+        &self,
+        pool: &PoolLedger<Balance>,
+    ) -> WithdrawalAmount<Balance> {
+        (
+            self.total_unstake_amount,
+            Rate::saturating_from_rational(self.total_stake_amount, pool.total_stake_amount)
```
GopherJ, line 114: here we should consider exchange rate as well so that after accumulating we dont need to use exchange rate

pallets/liquid-staking-v2/src/types.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    // after matching mechanism，for bond operation, user who unstake can get all amount directly
+    // and user who stake only get the matching part
+    pub fn instant_withdrawal_by_bond(
+        &self,
+        pool: &PoolLedger<Balance>,
+    ) -> WithdrawalAmount<Balance> {
+        (
+            self.total_unstake_amount,
```
GopherJ, line 113: here if we image in one era we only have one person stake, if exchangeRate = 2

Alice stakes 101KSM and unstakes 50xKSM, then alice is supposed to get:

100KSM & 50xKSM from the calculations but does it make sense?

pallets/liquid-staking-v2/src/types.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    // after matching mechanism，for unbond operation, user who stake can get all amount directly
+    // and user who unstake only get the matching part
+    pub fn instant_withdrawal_by_unbond(
+        &self,
+        pool: &PoolLedger<Balance>,
+    ) -> WithdrawalAmount<Balance> {
+        (
+            Rate::saturating_from_rational(self.total_unstake_amount, pool.total_unstake_amount)
+                .saturating_mul_int(pool.total_stake_amount),
```
GopherJ, line 127: 100KSM 50.5xKSM exchangeRate = 2

(100KSM, 100KSM) -> (100KSM, 50xKSM)

pallets/liquid-staking-v2/src/types.rs
GopherJ, line 113: 101KSM 50xKSM exchangeRate = 2

(100KSM, 100KSM) -> (100KSM, 50xKSM) -> bond1KSM

pallets/liquid-staking-v2/src/types.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub total_unstake_amount: Balance,
+    pub total_stake_amount: Balance,
+    pub operation_type: Option<StakingOperationType>,
+}
+
+impl<Balance> PoolLedger<Balance>
+where
+    Balance: AtLeast32BitUnsigned + FullCodec + Copy + MaybeSerializeDeserialize + Debug + Default,
+{
+    pub fn todo_after_new_era(&self) -> (StakingOperationType, Balance) {
```
GopherJ, line 47: actually I suggest this should be:

`todo_before_new_era` so that if stake-client died for some eras, it can call first todo_before_new_era, if it succeeds then it can update the era if not it shouldn't

pallets/liquid-staking-v2/src/types.rs
```diff
@@ -0,0 +1,131 @@
...
+        if self.total_stake_amount > self.total_unstake_amount {
+            (
+                StakingOperationType::Bond,
+                self.total_stake_amount - self.total_unstake_amount,
+            )
+        } else if self.total_stake_amount < self.total_unstake_amount {
+            (
+                StakingOperationType::Unbond,
+                self.total_unstake_amount - self.total_stake_amount,
```
GopherJ, line 56: total_unstake_amount should be clarified that it uses `KSM` as unit

MrPai:
will close this pr according to new design:
#384 
#385 
#386

PR #388, opened 2021-08-18
Record era index in realychain

### LINKED ISSUE #387: Allow stake-client push era_index
It needs to record era index in parachain due to current staking logic.

alannotnerd:
Part of liquidStakingV2.

Allow stake-client to submit era index mutation.

Close #387

### DIFF
--- a/pallets/liquid-staking-v2/src/lib.rs
+++ b/pallets/liquid-staking-v2/src/lib.rs
@@ -97,14 +97,16 @@ mod pallet {
         /// Set era index. Usually happend when era advanced in relaychain.
         #[pallet::weight(<T as Config>::WeightInfo::set_era_index())]
         #[transactional]
-        pub fn set_era_index(
+        pub fn trigger_new_era(
             origin: OriginFor<T>,
             era_index: EraIndex,
         ) -> DispatchResultWithPostInfo {
             let _who = ensure_signed(origin)?;
-            let (_, current_era_index) = Self::era_index_pair();
+            let current_era_index = Self::current_era();
             ensure!(current_era_index < era_index, Error::<T>::EraAlreadyPushed,);
-            EraIndexPair::<T>::put((Some(current_era_index), era_index));
+
+            PreviousEra::<T>::put(current_era_index);
+            CurrentEra::<T>::put(era_index);
             Self::deposit_event(Event::<T>::EraIndexUpdated(current_era_index, era_index));
             Ok(().into())
         }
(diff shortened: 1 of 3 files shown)


pallets/liquid-staking-v2/src/lib.rs
```diff
@@ -124,6 +151,13 @@ mod pallet {
         BalanceOf<T>,
     >;
 
+    /// Pair of (last_era_index, current_era_index).
+    ///
+    /// Note: In case of offchain-worker crashed, the last_era_index should also be recorded.
+    #[pallet::storage]
+    #[pallet::getter(fn era_index_pair)]
+    pub type EraIndexPair<T: Config> = StorageValue<_, (Option<EraIndex>, EraIndex), ValueQuery>;
```
GopherJ, line 159: we should store things in different storage because sometimes we just need one of them, also we will never have `None` case, if stake-client crashes it'll try to reapply eras

pallets/liquid-staking-v2/src/lib.rs
```diff
@@ -79,10 +88,28 @@ mod pallet {
...
+        /// Era has been pushed before.
+        EraAlreadyPushed,
     }
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
+        /// Set era index. Usually happend when era advanced in relaychain.
+        #[pallet::weight(<T as Config>::WeightInfo::set_era_index())]
+        #[transactional]
+        pub fn set_era_index(
```
GopherJ, line 100: it's called `trigger_new_era` from the spec

pallets/liquid-staking-v2/src/lib.rs
alannotnerd: I thought It doesn't make sense to have a default value for `previous_era` before we first update the era index. But it's okay to do so,  cause we don't read that value before updating. On the other hand, since they're always updated at the same time, It might be better to use a pair to record them.

pallets/liquid-staking-v2/src/lib.rs
alannotnerd: Fix it in next commit.

pallets/liquid-staking-v2/src/lib.rs
GopherJ: but when we change era we just need `current_era`

GopherJ: when we stake/unstake we just need `current_era`

alannotnerd: Fix it in next commit.

GopherJ: I think we should prefer having more meaningful names/functions

GopherJ (approved):

Issue #400, opened 2021-08-19
Implement matching bond/unbond functionality


yz89:
## Scopes
- [x] match the amount of bond/unbond in the current era by using `MatchingLedger`.
- [x] if the result is `bond`, emit `Bond` event, then transfer token to relaychain.
- [x] if the result is `unbond`, emit `Unbond` event. change `UnbondingOperationHistory` to ready.
- [x] unit test

GopherJ:
#405

crystalynliu:
We have a new idea for this one, #432

Issue #386, opened 2021-08-18
Implement claim dispatchable for staking v2


yz89:
## Scopes
- [x] add claim call with parameters accountId, ear_index
- [x] calculate the claimable amount. `UnstakeMisc.total_amount / total_unstake_amount * (total_unstake_amount - total_stake_amount)`.
- [x] transfer the claimable amount from pelletId to sender.
- [x] update the UnstakeMisc.claimed_amount
- [ ] unit test

MrPai:
This `claim` call relies on the Matching result, because after the matching mechanism:
for `bond` operation, users who `unstake` can claim all KSM back;
for `unbond` operation, users who `unstake` can only claim partial KSM back.

so we should first know how the matching mechanism store result.

yz89:
The matching result needn't store, for we can easily figure out it.
Maybe we can reuse the `matching` method implementing by @GopherJ, then we can calculate the claimable amount by matching result.

GopherJ:
#405

crystalynliu:
We have a new idea, #432

PR #405, opened 2021-08-19
finish matching & claim call, add ump transfer

GopherJ:
close: #400 #386

### DIFF
--- a/pallets/liquid-staking-v2/src/lib.rs
+++ b/pallets/liquid-staking-v2/src/lib.rs
@@ -38,17 +38,16 @@ mod pallet {
         traits::{Get, IsType},
         transactional, PalletId, Twox64Concat,
     };
-    use frame_system::{
-        ensure_signed,
-        pallet_prelude::{BlockNumberFor, OriginFor},
-    };
-    use orml_traits::{MultiCurrency, MultiCurrencyExtended};
+    use frame_system::{ensure_signed, pallet_prelude::*};
+    use orml_traits::{MultiCurrency, MultiCurrencyExtended, XcmTransfer};
     use sp_runtime::{traits::AccountIdConversion, ArithmeticError, FixedPointNumber};
+    use xcm::v0::MultiLocation;
 
     use primitives::{Amount, Balance, CurrencyId, EraIndex, Rate};
 
     use crate::types::{
-        MatchingLedger, Operation, ResponseStatus, StakingSettlementKind, UnstakeMisc,
+        MatchingLedger, Operation, OperationSatus, StakingOperationType, StakingSettlementKind,
+        UnstakeMisc,
     };
     use crate::weights::WeightInfo;
 
@@ -86,7 +85,18 @@ mod pallet {
         #[pallet::constant]
         type PalletId: Get<PalletId>;
 
+        /// Weight information for extrinsics in this pallet.
         type WeightInfo: WeightInfo;
+
+        /// XCM transfer provider
+        type XcmTransfer: XcmTransfer<Self::AccountId, BalanceOf<Self>, CurrencyId>;
+
+        /// Relaychain agent account
+        #[pallet::constant]
+        type RelayAgentLocation: Get<MultiLocation>;
+
+        /// Base xcm weight to use for cross chain transfer
+        type BaseXcmWeight: Get<Weight>;
     }
 
     #[pallet::event]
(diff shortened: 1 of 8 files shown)


pallets/liquid-staking-v2/src/tests.rs
```diff
@@ -160,7 +171,51 @@ fn test_record_unbond_response() {
...
+            MatchingLedger {
+                total_stake_amount: 10,
+                total_unstake_amount: 5,
+            }
+        );
+
+        assert_ok!(LiquidStaking::trigger_new_era(Origin::signed(Alice), 1));
+
+        assert_eq!(
+            StakingOperationHistory::<Test>::get(0, StakingOperationType::Bond),
```
yz89, line 198: replace 0 to `current_era`

pallets/liquid-staking-v2/src/tests.rs
```diff
@@ -160,7 +171,51 @@ fn test_record_unbond_response() {
...
+                block_number: 0_u64,
+                amount: 5
+            })
+        );
+
+        System::set_block_number(1);
+
+        assert_ok!(LiquidStaking::unstake(Origin::signed(Alice), 5));
+
+        assert_ok!(LiquidStaking::trigger_new_era(Origin::signed(Alice), 2));
```
yz89, line 210: replace 2 to `current_ear + 1`

pallets/liquid-staking-v2/src/types.rs
```diff
@@ -34,10 +20,60 @@ pub struct UnstakeMisc<Balance> {
...
+pub enum OperationSatus {
+    Pending,
+    Succeeded,
+    Failed,
+}
+
+impl MatchingLedger {
+    /// Before setting new era, ledger needs to check if bond/unbond should be done
+    /// on relaychain
+    pub fn summary(&self) -> (StakingOperationType, OperationSatus, Balance) {
```
yz89, line 59: Please add a unit test for `summary`.

pallets/liquid-staking-v2/src/types.rs
```diff
@@ -34,10 +20,60 @@ pub struct UnstakeMisc<Balance> {
 
 /// The matching pool's total stake & unstake amount in one era
 #[derive(Copy, Clone, Eq, PartialEq, Default, Encode, Decode, RuntimeDebug)]
-pub struct MatchingLedger<Balance> {
+pub struct MatchingLedger {
```
yz89, line 23: Is it necessary?

pallets/liquid-staking-v2/src/lib.rs
```diff
@@ -149,32 +182,18 @@ mod pallet {
...
-    pub type UnbondingOperationHistory<T: Config> =
-        StorageMap<_, Twox64Concat, EraIndex, Operation<BlockNumberFor<T>, BalanceOf<T>>>;
-
-    /// Record all the pending unstaking requests.
-    /// Key is the owner of assets.
-    #[pallet::storage]
-    #[pallet::getter(fn account_unstake)]
-    pub type AccountUnstake<T: Config> = StorageDoubleMap<
+    #[pallet::getter(fn staking_operation_history)]
+    pub type StakingOperationHistory<T: Config> = StorageDoubleMap<
```
yz89, line 188: This means we should record bond/unbond response either, but we are not going to add `record_bond` and `record_unbond` calls. So what do you think about that?

pallets/liquid-staking-v2/src/types.rs
GopherJ: without this I may need to add many restrictions for MathingLedger to make it supports `-` `+` operator, I think these restrictions are not necessary for now

pallets/liquid-staking-v2/src/lib.rs
GopherJ: this can be useful because sometimes we may need to check if `bond/unbond` is already done on relaychain, this status can be useful I think, so I think maybe we should have `record_status` call etc to record them

pallets/liquid-staking-v2/src/lib.rs
```diff
@@ -319,13 +416,29 @@ mod pallet {
             era_index: EraIndex,
         ) -> DispatchResultWithPostInfo {
             T::BridgeOrigin::ensure_origin(origin)?;
-            let current_era_index = Self::current_era();
-            ensure!(current_era_index < era_index, Error::<T>::EraAlreadyPushed,);
+            let current_era = Self::current_era();
+            ensure!(current_era < era_index, Error::<T>::EraAlreadyPushed,);
+
+            let (op_type, status, amount) = Self::matching_pool(current_era).summary();
```
MrPai, line 422: if we move this logic after `CurrentEra::<T>::put(era_index);`, and use `Self::matching_pool(previous_era)`, does it effect the logic, can you explain the different?

GopherJ: @MrPai it should be the same actually. If we put them in the same transactional call. But in previous version `trigger_new_era` only contain era update logic. I've put them together, not sure which one is the best because there are pros & cons

pallets/liquid-staking-v2/src/lib.rs
```diff
@@ -311,6 +339,75 @@ mod pallet {
...
+            if unstake_misc.total_amount.is_zero()
+                || unstake_misc.claimed_amount >= unstake_misc.total_amount
+            {
+                return Err(Error::<T>::NothingToClaim.into());
+            }
+
+            let current_era = Self::current_era();
+
+            // TODO check if asset has been transferred back from relaychain
+            // passed bonding duration & assets has been withdrawn to pallet account
```
yz89, line 367: You check the withdraw unbond state in `StakingOperationHistory` to address this TODO.

GopherJ:
close in favor of #431

Issue #339, opened 2021-08-01
account is reaped after calling liquidStaking.stake & liquidStaking.withdraw


GopherJ:
![image](https://user-images.githubusercontent.com/33961674/127773670-44d8845a-ca1b-4720-936d-8436860927bf.png)

alannotnerd:
I'll take care of this.

PR #407, opened 2021-08-20
add orml_xcm to allow xcm transact from technical commitee

GopherJ:
close: #339

as well

### DIFF
--- a/runtime/heiko/src/lib.rs
+++ b/runtime/heiko/src/lib.rs
@@ -25,7 +25,7 @@ mod weights;
 use codec::Encode;
 use frame_support::{
     dispatch::Weight,
-    traits::{All, LockIdentifier, U128CurrencyToVote},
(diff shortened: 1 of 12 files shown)


runtime/heiko/src/lib.rs
```diff
@@ -257,6 +269,12 @@ impl orml_tokens::Config for Runtime {
     type WeightInfo = ();
     type ExistentialDeposits = ExistentialDeposits;
     type MaxLocks = MaxLocks;
+    type DustRemovalWhitelist = DustRemovalWhitelist;
```
MrPai, line 272: Is it a new functionality in `orml_token`, can you explain how it works?

GopherJ: yes it's new functionality

GopherJ: @MrPai https://github.com/open-web3-stack/open-runtime-module-library/blob/master/tokens/src/lib.rs#L792

PR #450, opened 2021-08-24
Modify AMM pallet

### LINKED ISSUE #413: Modification in AMM Pallet
As a part of AMM Implementation, In phase 1

- modify the way we store and identify pools in storage to allow for a more efficient implementation
- supporting liquidity providing

cc @ETeissonniere @yubo-ruan

ashi20192020ayush:
resolves #413

### DIFF
--- a/pallets/amm/src/lib.rs
+++ b/pallets/amm/src/lib.rs
@@ -20,25 +20,29 @@
 
 extern crate alloc;
 
+#[cfg(test)]
+mod mock;
 mod pool_structs;
+
+#[cfg(test)]
+mod tests;
+
 use core::marker::PhantomData;
 use frame_support::pallet_prelude::*;
 use frame_support::{
     dispatch::DispatchResult,
     pallet_prelude::{StorageDoubleMap, StorageValue, ValueQuery},
-    traits::{EnsureOrigin, GenesisBuild, Get, Hooks, IsType},
-    Blake2_128Concat, PalletId, Parameter, Twox64Concat,
+    traits::{GenesisBuild, Get, Hooks, IsType},
+    Blake2_128Concat, PalletId, Twox64Concat,
 };
 use frame_system::ensure_signed;
 use frame_system::pallet_prelude::OriginFor;
 use orml_traits::{MultiCurrency, MultiCurrencyExtended};
 pub use pallet::*;
-use parallel_primitives::{Amount, Balance, CurrencyId, Rate};
-use pool_structs::{AMMCurve, PoolLiquidityAmount, StabilityPool};
-use sp_arithmetic::traits::BaseArithmetic;
+use pool_structs::PoolLiquidityAmount;
+use primitives::{Amount, Balance, CurrencyId, Rate};
 use sp_runtime::traits::AccountIdConversion;
 use sp_runtime::ArithmeticError;
-use sp_runtime::Perbill;
 
 #[frame_support::pallet]
 pub mod pallet {
@@ -57,22 +61,8 @@ pub mod pallet {
             Amount = Amount,
         >;
 
-        type Curve: AMMCurve;
-
-        type LpFee: Get<Perbill>;
-
         #[pallet::constant]
         type PalletId: Get<PalletId>;
-
-        type PoolId: Default + BaseArithmetic + Parameter;
-
-        type PoolManager: EnsureOrigin<Self::Origin>;
-
-        type StabilityPool: StabilityPool;
-
-        type TreasuryAccount: Get<Self::AccountId>;
-
-        type TreasuryFee: Get<Perbill>;
     }
 
     #[pallet::error]
@@ -138,8 +128,8 @@ pub mod pallet {
             NMapKey<Blake2_128Concat, CurrencyId>,
         ),
         PoolLiquidityAmount,
-		ValueQuery,
-		GetDefault,
+        ValueQuery,
+        GetDefault,
     >;
 
     /// A bag of liquidity composed by two different assets
@@ -178,6 +168,11 @@ pub mod pallet {
             ensure!(
                 !Pools::<T, I>::contains_key(base_asset, &quote_asset),
                 Error::<T, I>::PoolAlreadyExists
+            );
+
+            ensure!(
+                !LiquidityProviders::<T, I>::contains_key((who.clone(), base_asset, &quote_asset)),
+                Error::<T, I>::LiquidityProviderAlreadyExists
             );
 
             let amm_pool = PoolLiquidityAmount {
--- a/pallets/amm/src/lib.rs
+++ b/pallets/amm/src/lib.rs
@@ -32,7 +32,7 @@ mod pallet {
     use frame_support::{
         pallet_prelude::{StorageDoubleMap, StorageMap, StorageValue, ValueQuery},
         traits::{EnsureOrigin, GenesisBuild, Get, Hooks, IsType},
-        Blake2_128Concat, PalletId, Parameter,
+        Blake2_128Concat, PalletId, Parameter, Twox64Concat,
     };
     use orml_traits::MultiCurrencyExtended;
     use parallel_primitives::{Amount, Balance, CurrencyId, Rate};
@@ -115,9 +115,5 @@ mod pallet {
     /// A bag of liquidity composed by two different assets
     #[pallet::storage]
     pub type Pools<T: Config<I>, I: 'static = ()> =
-        StorageMap<_, Blake2_128Concat, T::PoolId, Pool>;
-
-    /// Auxiliary storage used to track pool ids
-    #[pallet::storage]
-    pub type PoolsCounter<T: Config<I>, I: 'static = ()> = StorageValue<_, T::PoolId, ValueQuery>;
+        StorageDoubleMap<_, Twox64Concat, CurrencyId, Twox64Concat, CurrencyId, Pool>;
 }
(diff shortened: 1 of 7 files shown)


pallets/amm/src/lib.rs
```diff
@@ -103,21 +119,149 @@ mod pallet {
...
+
+            let base_asset = sorted_assets.0;
+            let quote_asset = sorted_assets.1;
+            let base_amount = liquidity_amounts.0;
+            let quote_amount = liquidity_amounts.1;
+
+            ensure!(
+                !Pools::<T, I>::contains_key(base_asset, &quote_asset),
+                Error::<T, I>::PoolAlreadyExists
+            );
```
ETeissonniere, line 171: If I am right, this check would prevent future users to deposit more liquidity into the pool

pallets/amm/src/lib.rs
```diff
@@ -103,21 +119,149 @@ mod pallet {
...
+
+            ensure!(
+                !LiquidityProviders::<T, I>::contains_key((who.clone(), base_asset, &quote_asset)),
+                Error::<T, I>::LiquidityProviderAlreadyExists
+            );
```
ETeissonniere, line 176: If the liquidity provider already exists we should be able to increase their liquidity

pallets/amm/src/lib.rs
```diff
@@ -103,21 +119,149 @@ mod pallet {
...
+            Self::deposit_event(Event::<T, I>::LiquidityAdded(who, base_asset, quote_asset));
+            Ok(().into())
+        }
+
+        /// Allow users to remove liquidity from a given pool
+        ///
+        /// - `pool`: Currency pool, in which liquidity will be removed
+        /// - `liquidity_amounts`: Liquidity amounts to be removed from pool
+        #[pallet::weight(10_000)]
+        pub fn remove_liquidity(
```
ETeissonniere, line 197: Let's apply the [`#[transactional]`](https://substrate.dev/rustdocs/latest/frame_support/attr.transactional.html) modifier. Otherwise I believe there is a way to decrease the amount in the pool while making line 226 fail

pallets/amm/src/lib.rs
```diff
@@ -103,21 +119,149 @@ mod pallet {
...
-    #[pallet::storage]
-    pub type PoolsCounter<T: Config<I>, I: 'static = ()> = StorageValue<_, T::PoolId, ValueQuery>;
+    #[pallet::call]
+    impl<T: Config<I>, I: 'static> Pallet<T, I> {
+        /// Allow users to add liquidity to a given pool
+        ///
+        /// - `pool`: Currency pool, in which liquidity will be added
+        /// - `liquidity_amounts`: Liquidity amounts to be added in pool
+        #[pallet::weight(10_000)]
+        pub fn add_liquidity(
```
ETeissonniere, line 155:
```suggestion
        #[pallet::weight(10_000)]
        #[transactional]
        pub fn add_liquidity(
```

pallets/amm/src/lib.rs
ETeissonniere:
```suggestion
        #[pallet::weight(10_000)]
        #[transactional]
        pub fn remove_liquidity(
```

pallets/amm/src/lib.rs
```diff
@@ -103,21 +119,149 @@ mod pallet {
...
+                    pool_liquidity_amount.quote_amount = pool_liquidity_amount
+                        .quote_amount
+                        .checked_sub(quote_amount)
+                        .ok_or(ArithmeticError::Underflow)?;
+                    Ok(())
+                },
+            )?;
+
+            T::Currency::transfer(base_asset, &Self::account_id(), &who, base_amount)?;
+            T::Currency::transfer(quote_asset, &Self::account_id(), &who, quote_amount)?;
```
ETeissonniere, line 242: We actually don't withdraw `base_amount` and `quote_amount` from the pool. The amount sent back should be proportional to the amount of ownership of the pool by `origin`

ETeissonniere (changes requested):

pallets/amm/src/lib.rs
drbh: I agree this check fill throw an error if a user attempts to add liquidity into a pool. This is also supported by the test case here: https://github.com/parallel-finance/parallel/blob/c2348f798a87d761d963438581b193a0c8bcc48e/pallets/amm/src/tests.rs#L39

pallets/amm/src/lib.rs
drbh: A liquidity provider should be able to deposit tokens more than once, as stated above; the liquidity provider's balance should increase

pallets/amm/src/lib.rs
```diff
@@ -103,21 +119,149 @@ mod pallet {
...
+
+        /// Allow users to remove liquidity from a given pool
+        ///
+        /// - `pool`: Currency pool, in which liquidity will be removed
+        /// - `liquidity_amounts`: Liquidity amounts to be removed from pool
+        #[pallet::weight(10_000)]
+        pub fn remove_liquidity(
+            origin: OriginFor<T>,
+            pool: (CurrencyId, CurrencyId),
+            liquidity_amounts: (Balance, Balance),
```
drbh, line 200: Users deposit tokens and are entitled to a proportional share of the pool. They do not withdraw nominal amounts from the liquidity pool. Instead they should specify the proportion of their share they'd like to withdraw.

pallets/amm/src/tests.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        assert_noop!(
+            AMM::add_liquidity(Origin::signed(1.into()), (DOT, XDOT), (10, 20)),
+            Error::<Test>::PoolAlreadyExists,
+        );
+    })
+}
+
+#[test]
+fn remove_liquidity_should_work() {
```
drbh, line 46: This test should account for mores cases

1. A pool with a single LP provider who deposit tokens and withdraws their whole share (most simple case)
2. A pool with a single LP provider who deposit tokens and withdraws a portion of their total shares (simple case)
3. A pool with multiple LP providers who deposit tokens with varying ratios and some withdraw their shares (complex example)

pallets/amm/src/lib.rs
```diff
@@ -103,21 +119,149 @@ mod pallet {
...
+        #[pallet::weight(10_000)]
+        pub fn add_liquidity(
+            origin: OriginFor<T>,
+            pool: (CurrencyId, CurrencyId),
+            liquidity_amounts: (Balance, Balance),
+        ) -> DispatchResult {
+            let who = ensure_signed(origin)?;
+            let sorted_assets = &Self::get_upper_currency(pool.0, pool.1);
+
+            let base_asset = sorted_assets.0;
```
GopherJ, line 163: this looks ugly, you can expand it using `let (a, b)`

pallets/amm/src/lib.rs
```diff
@@ -103,21 +119,149 @@ mod pallet {
...
+            let quote_asset = sorted_assets.1;
+            let base_amount = liquidity_amounts.0;
```
GopherJ, line 207: this looks ugly

pallets/amm/src/mock.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+impl From<u64> for AccountId {
+    fn from(account_id: u64) -> Self {
+        Self(account_id)
+    }
+}
+
+impl From<AccountId> for [u8; 32] {
```
GopherJ, line 57: we only need this in liquid staking mock because we've used XcmTransfer

pallets/amm/src/lib.rs
ashi20192020ayush: @ETeissonniere Could you please suggest how should I calculate this?

pallets/amm/src/lib.rs
ashi20192020ayush: @drbh Are you saying that I will get withdraw amount in parameter

ETeissonniere: maybe we could do something like this:
```suggestion
            ownership: Perbill,
```
If `ownership` is 100% we remove all their liquidity, if 50% we remove half of the liquidity they own, etc...

pallets/amm/src/lib.rs
ETeissonniere: https://www.notion.so/parallelfinance/Ownership-and-Liquidity-9ae44353a5f24fe6bd02640cf41ed0af

ETeissonniere: https://www.notion.so/parallelfinance/Ownership-and-Liquidity-9ae44353a5f24fe6bd02640cf41ed0af

pallets/amm/src/lib.rs
ETeissonniere: See https://www.notion.so/parallelfinance/Ownership-and-Liquidity-9ae44353a5f24fe6bd02640cf41ed0af

pallets/amm/src/lib.rs
```diff
@@ -103,21 +115,201 @@ mod pallet {
...
+                            .checked_add(base_amount)
+                            .ok_or(ArithmeticError::Overflow)?;
+                        pool_liquidity_amount.quote_amount = pool_liquidity_amount
+                            .quote_amount
+                            .checked_add(quote_amount)
+                            .ok_or(ArithmeticError::Overflow)?;
+                        pool_liquidity_amount.ownership = ownership;
+                        Ok(())
+                    },
+                )?;
```
ETeissonniere, line 188: `contains_key` + `Self::pools` + `try_mutate` will generate 3 reads and 1 write. We can optimize this via `try_mutate_exists`

ETeissonniere (commented):
amazing! My only comment would be that we may have to think of optimizing DB reads and writes

PR #481, opened 2021-08-27
Added benchmarking for AMM pallet

### LINKED ISSUE #458: Write benchmarks for AMM pallet

ashi20192020ayush:
resolves #458


pallets/amm/src/lib.rs
```diff
@@ -128,7 +134,18 @@ pub mod pallet {
         ///
         /// - `pool`: Currency pool, in which liquidity will be added
         /// - `liquidity_amounts`: Liquidity amounts to be added in pool
-        #[pallet::weight(10_000)]
+        #[pallet::weight(
+		{
+			let p = *pool;
+			let (_, base_asset, quote_asset) =
+				Pallet::<T, I>::get_upper_currency(p.0, p.1);
+			if Pools::< T, I >::contains_key(base_asset, quote_asset) {
```
ETeissonniere, line 142: This means we would have to one read to the computed weights. An alternative implementation would be to:
1. Default to the highest weight of both
2. When the extrinsic executes return the actual weight we want to charge the user for

ETeissonniere: Or we can keep it simple and just add one read

ETeissonniere (approved):
Lgtm

PR #485, opened 2021-08-28
fix system.setCode's Digest item must match calculated

GopherJ:
according to https://github.com/paritytech/cumulus/issues/474, we need `#[sp_version::runtime_version]` to make system.setCode works

this also turns on `archive` mode for collator


GopherJ:
need to test locally runtime upgrade again from when there is no `sp_version::runtime_version` macro

PR #494, opened 2021-08-30
Added Liquidity Ratio Support

### LINKED ISSUE #471: Liquidity Ratio Support
Adding liquidity to an AMM pool should optimize the deposit's ratio to match the current exchange rate. Otherwise, people could game the current pool exchange rate.

ashi20192020ayush:
resolves #471

### DIFF
--- a/pallets/amm/src/lib.rs
+++ b/pallets/amm/src/lib.rs
@@ -177,9 +177,10 @@ pub mod pallet {
                         (optimal_base_amount, quote_amount)
                     };
 
-                let (minimum_base_amount, minimum_quote_amount) = match is_inverted {
-                    true => (minimum_amounts.1, minimum_amounts.0),
-                    false => (minimum_amounts.0, minimum_amounts.1),
+                let (minimum_base_amount, minimum_quote_amount) = if is_inverted {
+                    (minimum_amounts.1, minimum_amounts.0)
+                } else {
+                    (minimum_amounts.0, minimum_amounts.1)
                 };
 
                 ensure!(
--- a/pallets/amm/src/lib.rs
+++ b/pallets/amm/src/lib.rs
@@ -180,10 +180,10 @@ pub mod pallet {
                 let (minimum_base_amount, minimum_quote_amount) = minimum_amounts;
 
                 ensure!(
-                    ideal_base_amount > minimum_base_amount
-                        && ideal_quote_amount > minimum_quote_amount
-                        && ideal_base_amount < base_amount
-                        && ideal_quote_amount < quote_amount,
+                    ideal_base_amount >= minimum_base_amount
+                        && ideal_quote_amount >= minimum_quote_amount
+                        && ideal_base_amount <= base_amount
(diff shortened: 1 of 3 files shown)


pallets/amm/src/lib.rs
```diff
@@ -156,6 +159,38 @@ pub mod pallet {
...
+
+                let (ideal_base_amount, ideal_quote_amount): (Balance, Balance) =
+                    if optimal_quote_amount <= quote_amount {
+                        (base_amount, optimal_quote_amount)
+                    } else {
+                        let optimal_base_amount = Self::quote(
+                            quote_amount,
+                            pool_liquidity_amount.base_amount,
+                            pool_liquidity_amount.quote_amount,
+                        );
```
drbh, line 176: I think the base and quote amounts need to be inverse from the `optimal_quote_amount` in order to calculate `optimal_base_amount`


```rust
let optimal_base_amount = Self::quote(
    quote_amount,
    pool_liquidity_amount.quote_amount,
    pool_liquidity_amount.base_amount,
);
```

[similar to uniswap](https://github.com/Uniswap/uniswap-v2-periphery/blob/dda62473e2da448bc9cb8f4514dadda4aeede5f4/contracts/UniswapV2Router02.sol#L54)

drbh (changes requested):

pallets/amm/src/lib.rs
ETeissonniere: I second that, this seems correct

pallets/amm/src/lib.rs
```diff
@@ -156,6 +159,38 @@ pub mod pallet {
...
+                        (optimal_base_amount, quote_amount)
+                    };
+
+                let (minimum_base_amount, minimum_quote_amount) = match is_inverted {
+                    true => (minimum_amounts.1, minimum_amounts.0),
+                    false => (minimum_amounts.0, minimum_amounts.1),
+                };
```
ETeissonniere, line 183: I think this can actually be rewritten to this
```suggestion
                let (minimum_base_amount, minimum_quote_amount) = if is_inverted {
                    (minimum_amounts.1, minimum_amounts.0)
                } else {
                    (minimum_amounts.0, minimum_amounts.1)
                };
```

drbh:
this looks good! However, I think @ETeissonniere mentioned adding a unit test "to ensure / demo that we do select the correct ratio". Should we add that before merging?

ashi20192020ayush:
> this looks good! However, I think @ETeissonniere mentioned adding a unit test "to ensure / demo that we do select the correct ratio". Should we add that before merging?

@ETeissonniere @drbh I think this is already covered in existing test-cases because I modified existing test-cases to execute successfully. I also added one negative scenario https://github.com/parallel-finance/parallel/blob/258716f7b6a4e515d60187899673c9b409335ec4/pallets/amm/src/tests.rs#L56

Please provide some details about new test-case scenario.

ETeissonniere (approved):

PR #506, opened 2021-09-01
restruct the loans pallet

yz89:


pallets/loans/src/interest.rs
```diff
@@ -0,0 +1,119 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use primitives::{Balance, CurrencyId};
+use sp_runtime::{traits::Zero, DispatchResult};
+
+use crate::*;
+
+impl<T: Config> Pallet<T> {
+    /// Accrue interest per block and update corresponding storage
+    pub(crate) fn accrue_interest() -> DispatchResult {
```
GopherJ, line 10: is `pub(crate)` necessary? is it used in test? I think maybe `fn` is already enough

pallets/loans/src/interest.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /// Update the borrow index by borrow rate, the total borrows and
+    /// total reserves will be updated simultaneously.
+    ///
+    /// interestAccumulated = totalBorrows * borrowRate
+    /// totalBorrows = interestAccumulated + totalBorrows
+    /// totalReserves = interestAccumulated * reserveFactor + totalReserves
+    /// borrowIndex = borrowIndex * (1 + borrowRate)
+    pub(crate) fn update_borrow_index(
```
GopherJ, line 87: maybe we can add https://substrate.dev/rustdocs/latest/frame_support/attr.require_transactional.html marker so that we are sure this function will only be used in transactional calls

pallets/loans/src/interest.rs
```diff
@@ -0,0 +1,119 @@
...
+            Self::update_exchange_rate(currency_id)?;
+        }
+
+        Ok(())
+    }
+
+    /// Update the exchange rate according to the totalCash, totalBorrows and totalSupply.
+    ///
+    /// exchangeRate = (totalCash + totalBorrows - totalReserves) / totalSupply
+    pub(crate) fn update_exchange_rate(currency_id: CurrencyId) -> DispatchResult {
```
GopherJ, line 38: maybe we can add https://substrate.dev/rustdocs/latest/frame_support/attr.require_transactional.html

pallets/loans/src/interest.rs
GopherJ, line 10: maybe we can add https://substrate.dev/rustdocs/latest/frame_support/attr.require_transactional.html

GopherJ:
otherwise LGTM

pallets/loans/src/interest.rs
yz89: Yes, otherwise we can not invoke it in lib.rs and test.rs.

pallets/loans/src/interest.rs
yz89: It will break the test because the function is not called within a storage transaction.
It will cause panic: 'Require transaction not called within with_transaction'.

GopherJ (approved):

PR #511, opened 2021-09-01
avoid the conflict of metadata between orml_vesting and build-in vesting

yz89:
Alias the orml_vesting to update the parallel-js.
https://github.com/parallel-finance/parallel-js/pull/3

### DIFF
--- a/runtime/heiko/src/lib.rs
+++ b/runtime/heiko/src/lib.rs
@@ -227,7 +227,7 @@ impl Contains<Call> for BaseCallFilter {
         // Call::Scheduler(_) |
 
         // // 3rd Party
-        // Call::Vesting(_) |
+        // Call::OrmlVesting(_) |
         // Call::Currencies(_) |
         // Call::Oracle(_) |
         // Call::XTokens(_) |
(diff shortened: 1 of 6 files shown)


runtime/heiko/src/lib.rs
```diff
@@ -1196,7 +1196,7 @@ construct_runtime!(
         XTokens: orml_xtokens::{Pallet, Storage, Call, Event<T>} = 43,
         UnknownTokens: orml_unknown_tokens::{Pallet, Storage, Event} = 44,
         OrmlXcm: orml_xcm::{Pallet, Call, Event<T>} = 45,
-        Vesting: orml_vesting::{Pallet, Storage, Call, Event<T>, Config<T>} = 46,
+        OrmlVesting: orml_vesting::{Pallet, Storage, Call, Event<T>, Config<T>} = 46,
```
GopherJ, line 1199: shall we do something on the underlying storage?

yz89:
Fixed by https://github.com/parallel-finance/parallel-js/pull/3/commits/8a365bef8711ca3b9a2e34d305566d62f960f2e9

PR #526, opened 2021-09-04
Feat/amm trade trait

drbh:
Hey guys, this PR implements the trade trait. It does not include protocol fees, since I have some questions about the `ProtocolFeeHandler`. Also I'm not sure if i've calculated LP fees correctly...

Since LP fees are calculated by 
```
let fees = T::LpFee::get() * amount_in;
```

If `amount_in` is less then `1/T::LpFee::get()`, the fee will be < 1, and is returned as 0. 

Is this the intended functionality of LP fees?

Please let me know what changes should be made


pallets/amm/src/lib.rs
```diff
@@ -74,6 +80,16 @@ pub mod pallet {
...
+
+        /// Defines the fees taken out of each trade and sent back to the AMM pool,
+        /// typically 0.3%.
+        type LpFee: Get<Perbill>;
+
+        /// How much the protocol is taking out of each trade.
+        type ProtocolFee: Get<Perbill>;
+
+        // -- TODO --
+        // type ProtocolFeeHandler: OnUnbalanced<CreditOf<Self, I>>;
```
ETeissonniere, line 92: Both ORML Tokens and Parity's Assets support the new `fungibles::*` traits from Substrate. We can use those to generate the `CreditOf` Imbalance and pass it to an `OnUnbalanced` runtime implementation

drbh: @ETeissonniere can you help me determine what type `ProtocolFeeHandler` should be? I've tried the following: 
```rust
type ProtocolFeeHandler: OnUnbalanced<CreditOf<Self::AccountId, Self>>;
```

but get this error:
```rust
error[E0277]: the trait bound `Self: hidden_include::traits::fungibles::Unbalanced<<Self as frame_system::Config>::AccountId>` is not satisfied
```
Thank you!

ETeissonniere: @drbh I think the last `Self` should actually be the tokens pallet

pallets/amm/src/lib.rs
```diff
@@ -455,4 +476,119 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
             .checked_div(base_pool)
             .expect("cannot overflow with positive divisor; qed")
     }
+
+    pub fn trade(
```
ETeissonniere, line 480: This would need to be moved to an implementation of the `AMM` trait

ETeissonniere (approved):

pallets/amm/src/lib.rs
```diff
@@ -465,3 +483,127 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+                // 6. Wire amount_in of the input token (identified by pair.0) from who to PalletId
+                T::Currency::transfer(
+                    pair.0,
+                    &who,
+                    &Self::account_id(),
+                    amount_without_protocol_fees,
+                )?;
+
+                // 7. Wire amount_out of the output token (identified by pair.1) to who from PalletId
+                T::Currency::transfer(pair.1, &Self::account_id(), &who, amount_out)?;
```
ashi20192020ayush, line 588:
```suggestion
                T::Currency::transfer(quote_asset, &Self::account_id(), &who, amount_out)?;
```

pallets/amm/src/lib.rs
```diff
@@ -465,3 +483,127 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+
+                // 8. Wire protocol fees as needed (input token)
+                T::Currency::transfer(pair.0, &who, &T::ProtocolFeeReceiver::get(), protocol_fees)?;
```
ashi20192020ayush, line 591:
```suggestion
                T::Currency::transfer(base_asset, &who, &T::ProtocolFeeReceiver::get(), protocol_fees)?;
```

pallets/amm/src/tests.rs
```diff
@@ -366,3 +366,164 @@ fn remove_liquidity_with_more_liquidity_should_not_work() {
...
+        let trader = AccountId::from(4_u64);
+
+        // create pool and add liquidity
+        AMM::add_liquidity(
+            Origin::signed(3.into()),
+            (DOT, XDOT),
+            (100_000_000, 100_000_000),
+            (99_999, 99_999),
+        )
+        .expect("Error initalizing AMM");
```
ashi20192020ayush, line 384: Why it would throw error?

ashi20192020ayush (changes requested):

pallets/amm/src/tests.rs
drbh: it shouldn't! 🙂 I'll update them to use `asset_ok!` like the earlier examples.

pallets/amm/src/lib.rs
drbh: I don't think that `pair.0` and `quote_asset` are always the same. 

We know that `pair.0` will be the input currency from the trade, where `quote_asset` is the RHS asset in our pool. It's possible that a trader will want to input the `quote_asset` of a pool and get the `base_asset` in return. 

Due to this we can not assume that `pair.0 === quote_asset`

pallets/amm/src/lib.rs
drbh: same reasoning as the comment above.  

Earlier in the `trade` function we use the following if statement: 

```rust
// supply_in == liquidity_amount.base_amount unless inverted
let (supply_in, supply_out) = if is_inverted {
    (liquidity_amount.quote_amount, liquidity_amount.base_amount)
} else {
    (liquidity_amount.base_amount, liquidity_amount.quote_amount)
};
```

The above statement handles the case where `pair.0 != base_asset`. Therefore we'll need access the values from the tuple directly

pallets/amm/src/lib.rs
ashi20192020ayush: Can we extract variables from pair? using pair.1 and pair.2 doesn't look good. I also used in the past but team didn't like this. Please see this https://github.com/parallel-finance/parallel/pull/450#discussion_r695339486

pallets/amm/src/tests.rs
ashi20192020ayush: please do the same for other test-cases as well

pallets/amm/src/lib.rs
drbh: yep thats a good idea 👌

ashi20192020ayush (approved):

PR #537, opened 2021-09-06
Feat/small improvements

GopherJ:


pallets/liquid-staking-v2/src/lib.rs
```diff
@@ -205,8 +216,9 @@ mod pallet {
         ///     - Pallet's balance is insufficiant.
         ///     - Queue is empty.
         ///     - `remaining_weight` is less than one pop_queue needed.
-        fn on_idle(_n: BlockNumberFor<T>, remaining_weight: Weight) -> Weight {
-            let mut remaining_weight = remaining_weight;
+        fn on_idle(_n: BlockNumberFor<T>, mut remaining_weight: Weight) -> Weight {
+            /// TODO should use T::WeightInfo::on_idle instead
```
alannotnerd, line 220: It stands the weight during a single loop. The total weight of this hook is unable to be calculated.

pallets/liquid-staking-v2/src/lib.rs
```diff
@@ -218,17 +230,14 @@ mod pallet {
                     break;
                 }
 
-                // get the front of the queue.
-                let (who, amount) = Self::unstake_queue()
-                    .first()
-                    .expect("Has been checked before.")
-                    .clone();
+                // Get the front of the queue.
+                let (who, amount) = &Self::unstake_queue()[0];
```
alannotnerd, line 234: `first` is a more meaningful statement. You could remove `clone`  if you just want to use reference.

pallets/liquid-staking-v2/src/lib.rs
GopherJ: I read acala's code, it seems that we shouldn't use loop inside `on_idle` because in this case it won't give other pallets the chance to execute stuff. So I'm wondering if we should only remove one unstake request at one time. If there aren't many pallets using `on_idle` then `pallet_liquid_staking_v2` will still have many chances to execute stuff

pallets/liquid-staking-v2/src/lib.rs
GopherJ: that will meet rust compiler issue

alannotnerd (approved):

pallets/liquid-staking-v2/src/lib.rs
GopherJ: yes let's think about it later

PR #539, opened 2021-09-06
Feat/liquid staking v3

GopherJ:


yz89:
We don't need to maintain more pallets at the same time. 'liquid-staking' should be considered final, supporting XCm-transact.

GopherJ:
> We don't need to maintain more pallets at the same time. 'liquid-staking' should be considered final, supporting XCm-transact.

so you want to reuse `liquid-staking v1` for testing xcm.transact? I thought we don't want to touch it

yz89:
Yes, you can delete all the code in `liquid-staking` and rename `liquid-staking-v3` to `liquid-staking`.

GopherJ:
> Yes, you can delete all the code in `liquid-staking` and rename `liquid-staking-v3` to `liquid-staking`.

ok I'll test runtime upgrade

MrPai:
> Yes, you can delete all the code in `liquid-staking` and rename `liquid-staking-v3` to `liquid-staking`.

Shall we keep on the job according to this PR, or raise a new PR?

PR #551, opened 2021-09-08
Feat/unbond/rebond/withdraw_unbonded transact

GopherJ:


GopherJ:
> Why allow dead_code?

because it's not used in pallet-liquid-staking yet

pallets/liquid-staking/src/tests.rs
```diff
@@ -219,20 +219,71 @@ fn test_transact_bond_extra_work() {
...
+        assert_eq!(ledger.active, 0u128);
+    });
+}
+
+#[test]
+fn test_transact_rebond_work() {
+    TestNet::reset();
+
+    ParaA::execute_with(|| {
+        assert_ok!(LiquidStaking::bond(
```
yz89, line 268: Could you please use the case: Bond 10, Unbond 5, Rebond 3 ?

pallets/liquid-staking/src/tests.rs
```diff
@@ -219,20 +219,71 @@ fn test_transact_bond_extra_work() {
...
         assert_eq!(ledger.total, 5 * DOT_DECIMAL);
     });
 }
+
+#[test]
+fn test_transact_unbond_work() {
+    TestNet::reset();
+
+    ParaA::execute_with(|| {
+        assert_ok!(LiquidStaking::bond(
```
yz89, line 241: Cloud you please use the case that Bond 5 and then Unbond 2? And ensure ledger.total = 5, ledger.active = 3.

pallets/liquid-staking/src/types.rs
```diff
@@ -91,7 +91,27 @@ pub struct StakingBondCall<T: Config> {
 pub struct StakingBondExtraCall<T: Config> {
```
yz89, line 91: There are too many constraints once using`<T: Config>`.  `<Balance>` is enough.  https://github.com/parallel-finance/parallel/blob/master/pallets/liquid-staking-v2/src/types.rs#L14

yz89:
> > Why allow dead_code?
> 
> because these functions are not used in pallet-liquid-staking yet, will be considered dead code

Got it.

pallets/liquid-staking/src/types.rs
GopherJ: Balance I think may requires some trait bounds to make sure it's a `number`

GopherJ: or codec

yz89: When you bond it with BalanceOf<T>, the Balance will be derived as `MultiCurrency<<T as frame_system::Config>::AccountId>>::Balance` which is u128.

https://github.com/parallel-finance/parallel/blob/master/pallets/liquid-staking-v2/src/lib.rs#L177

GopherJ: but what's the value of this change? `Balance` is even longer than `Config`

GopherJ: using https://github.com/parallel-finance/parallel/blob/master/pallets/liquid-staking-v2/src/types.rs#L14 I'll need to add trait bound

yz89: You needn't. 
The type Balance in MultiCurrency like this:
`type Balance: AtLeast32BitUnsigned + FullCodec + Copy + MaybeSerializeDeserialize + Debug + Default;`

AtLeast32BitUnsigned contains BaseArithmetic which already have the trait Bounded.
You can treat as number by using BaseArithmetic trait.

yz89 (approved):

pallets/liquid-staking/src/types.rs
GopherJ: maybe you can write some code here so that  I can better understand

```rust
/// Relaychain staking.unbond call arguments
#[derive(Clone, Encode, Decode, RuntimeDebug)]
pub struct StakingRebondCall<T: Config> {
    /// [pallet index, call index]
    pub call_index: [u8; 2],
    /// Rebond amount
    #[codec(compact)]
    pub value: T::Balance,
}
```
do you want to do this?

pallets/liquid-staking/src/bridge.rs
```diff
@@ -108,4 +110,112 @@ where
...
+            }
+            Err(_e) => {
+                return Err(Error::<T>::RebondCallFailed.into());
+            }
+        }
+        Ok(())
+    }
+
+    /// withdraw unbonded on relaychain via xcm.transact
+    pub(crate) fn withdraw_unbonded(num_slashing_spans: u32) -> DispatchResult {
```
MrPai, line 187: It seems like no test for this method, are we going to add a unit test for it?

GopherJ: yes I'm adding it

pallets/liquid-staking/src/types.rs
yz89:
```suggestion
/// Relaychain staking.unbond call arguments
#[derive(Clone, Encode, Decode, RuntimeDebug)]
pub struct StakingRebondCall<Balance> {
    /// [pallet index, call index]
    pub call_index: [u8; 2],
    /// Rebond amount
    #[codec(compact)]
    pub value: Balance,
}
```

PR #568, opened 2021-09-13
Add constant notation for `relayAgent`

alannotnerd:


runtime/vanilla/src/lib.rs
```diff
@@ -466,6 +466,7 @@ parameter_types! {
         }
     );
     pub const PeriodBasis: BlockNumber = 1000u32;
+    pub BridgeAccount: Vec<AccountId> = vec![hex!["306721211d5404bd9da88e0204360a1a9ab8b87c66c1bc2fcdd37f3c2222cc20"].into()];
```
GopherJ, line 469: can you add also this to parallel runtime?

runtime/vanilla/src/lib.rs
```diff
@@ -474,7 +475,7 @@ impl pallet_liquid_staking::Config for Runtime {
     type PalletId = StakingPalletId;
     type StakingCurrency = StakingCurrency;
     type LiquidCurrency = LiquidCurrency;
-    type BridgeOrigin = EnsureSigned<AccountId>;
+    type BridgeOrigin = EnsureSignedBy<IsInVec<BridgeAccount>, AccountId>;
```
GopherJ, line 478: this should also be in parallel runtime

pallets/liquid-staking-v2/src/lib.rs
```diff
@@ -100,6 +100,7 @@ mod pallet {
         type BaseXcmWeight: Get<Weight>;
 
         /// Account manages the staking assets.
+        #[pallet::constant]
```
GopherJ, line 103: BaseXcmWeight also needs pallet::constant, can you correct also `pallet-liquid-staking`?

GopherJ: maybe the new `DerivativeIndex` added by me needs also `constant` notation:)

GopherJ (approved):

PR #574, opened 2021-09-13
replace match with if

drbh:
small change so all `is_inverted` conditionals use the same style. **not pressing


GopherJ (commented):
hi, what is `pallets/heiko.json` for?

drbh:
@GopherJ I accidentally added that file. Just removed

GopherJ:
no problem, strange that `clippy` didn't warn about this `match`

GopherJ (approved):

PR #577, opened 2021-09-13
apply pallet-assets in loans pallet

### LINKED ISSUE #519: replace orml-tokens by using pallet-assets
## Scopes
- [x] add Assets trait for loans pallet
- [x] define AssetIdOf
- [x] replace orml-tokens in loans pallet
- [ ] fix test and benchmark

yz89:
close #519


pallets/liquidation/src/lib.rs
```diff
@@ -114,7 +124,11 @@ pub mod pallet {
     }
 
     #[pallet::hooks]
-    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T>
+    where
```
GopherJ, line 128: do we really need these two trait bounds here? because you didn't use in hooks

pallets/loans/src/lib.rs
```diff
@@ -343,7 +360,11 @@ pub mod pallet {
     pub struct Pallet<T>(PhantomData<T>);
 
     #[pallet::hooks]
-    impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
+    impl<T: Config> Hooks<T::BlockNumber> for Pallet<T>
+    where
```
GopherJ, line 364: we have too many such trait bounds

GopherJ (approved):

pallets/liquidation/src/lib.rs
yz89: Yes, we need for calling `accrue_interest` which is under the trait bounds.

ETeissonniere (commented):
Not requesting changes to make sure I don't block you guys.

I am concerned the changes in this PR would prevent us to use our native token for loans. In order to support this, the AMM team is developing a `CurrencyAdapter` in #584. I would suggest you guys use this as well.

GopherJ:
let's also remember to update changelog

yz89:
> Not requesting changes to make sure I don't block you guys.
> 
> I am concerned the changes in this PR would prevent us to use our native token for loans. In order to support this, the AMM team is developing a `CurrencyAdapter` in #584. I would suggest you guys use this as well.

Ok, we will use it when `CurrencyAdapter` is stable. I think there are other factors to consider when using the native token, such as tx fee will not be enough after deposit, thus the transaction will fail.

yz89:
> let's also remember to update changelog

sure

PR #584, opened 2021-09-14
Deprecate orml tokens

### LINKED ISSUE #544: Replace ORML tokens with pallet-assets

ashi20192020ayush:
resolves https://github.com/parallel-finance/workspace/issues/35


primitives/src/currency.rs
```diff
@@ -72,3 +72,31 @@ impl TryFrom<CurrencyId> for TokenSymbol {
...
+    Asset(u32),
+}
+
+#[derive(Encode, Decode, Eq, PartialEq, Copy, Clone, RuntimeDebug, PartialOrd, Ord)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize, Hash))]
+pub enum CurrencyOrAsset {
+    NativeCurrency,
+    Asset(u32),
+    PoolAsset(PoolAsset, PoolAsset),
+}
```
ETeissonniere, line 88: I don't get why you have to create `PoolAsset` since you could create an asset on the fly in `pallet-assets` which would then work with `CurrencyOrAsset::Asset`.

You'd have a flow like this in the event when you are creating a new pool:
1. Call `pallet-assets`'s `force_create` extrinsic to create a new asset from our own `AccountId`
2. Map our in storage pool state to the asset created

ETeissonniere (changes requested):
I am not sure we need the `PoolAssets` enum, ideally we'd just depend on the original enum.

primitives/src/currency.rs
ashi20192020ayush: @ETeissonniere `pallet-assets`'s `force_create` would work for one asset at a time. We would get two assets `base_asset` and `quote_asset` in our dispatch able functions. 


If I understand your point, we have to create two assets against `base_asset` and `quote_asset`. We would also call `total_issuance` `mint_into` `burn_from` twice for each asset. 

Please correct me if I am wrong. Lets have a call to discuss this.

ETeissonniere: Let's chat on our call. I think there are some confusions

pallets/amm/src/benchmarking.rs
```diff
@@ -10,21 +10,20 @@ use frame_benchmarking::{
 };
 use frame_support::assert_ok;
 use frame_system::{self, RawOrigin as SystemOrigin};
-use orml_traits::MultiCurrency;
 use primitives::TokenSymbol;
 use sp_std::prelude::*;
 
-const BASE_ASSET: CurrencyId = CurrencyId::Token(TokenSymbol::xDOT);
-const QUOTE_ASSET: CurrencyId = CurrencyId::Token(TokenSymbol::DOT);
+const BASE_ASSET: CurrencyOrAsset = CurrencyOrAsset::NativeCurrency(TokenSymbol::xDOT);
```
GopherJ, line 16: Can we change the name? `CurrencyOrAsset` is too ugly

ashi20192020ayush: @GopherJ  I added as per this notion page https://www.notion.so/parallelfinance/Currency-and-Assets-Adapter-b6c5995794574086b40079ae2947dce6

@ETeissonniere Could you please suggest some other name?

ETeissonniere: @GopherJ let us know which name you would prefer :)

primitives/src/currency.rs
```diff
@@ -161,3 +161,10 @@ impl<
         Ok(asset.clone().into())
     }
 }
+
+#[derive(Encode, Decode, Eq, PartialEq, Copy, Clone, RuntimeDebug, PartialOrd, Ord)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize, Hash))]
+pub enum CurrencyOrAsset {
+    NativeCurrency(TokenSymbol),
+    Asset(u32),
```
ETeissonniere, line 169: Actually we could even do this
```suggestion
    Asset(AssetId),
```

ETeissonniere (commented):
lgtm, let's remove the orml pallets from the `Cargo.toml`

ashi20192020ayush:
> lgtm, let's remove the orml pallets from the `Cargo.toml`

@ETeissonniere  I already removed from AMM pallet.

ETeissonniere (approved):

pallets/amm/src/mock.rs
```diff
@@ -1,25 +1,31 @@
...
     testing::Header,
     traits::{BlakeTwo256, IdentityLookup},
     RuntimeDebug,
 };
-
-pub const DOT: CurrencyId = CurrencyId::Token(TokenSymbol::DOT);
-pub const XDOT: CurrencyId = CurrencyId::Token(TokenSymbol::xDOT);
-pub const NATIVE: CurrencyId = CurrencyId::Token(TokenSymbol::HKO);
+use std::marker::PhantomData;
+pub const DOT: CurrencyOrAsset = CurrencyOrAsset::NativeCurrency(TokenSymbol::DOT);
```
yz89, line 26: Why native token is also a TokenSymbol::DOT ? I think a native token is like HKO, and DOT is a kind of asset management by pallet assets.

runtime/parallel/src/lib.rs
```diff
@@ -1195,11 +1203,115 @@ parameter_types! {
     pub DefaultProtocolFeeReceiver: AccountId = TreasuryPalletId::get().into_account();
 }
 
+pub struct Adapter<AccountId> {
```
yz89, line 1206: Could we add a supporting pallet like orml-currencies to handle the duplicate code?

pallets/amm/src/mock.rs
ashi20192020ayush: What do you suggest @ETeissonniere ? Shall we rename NativeCurrency or add one more field in `CurrencyOrAsset`

runtime/parallel/src/lib.rs
ashi20192020ayush: Sorry @yz89 I didn't understand this comment. Could you please explain more?

cc @ETeissonniere

pallets/amm/src/mock.rs
yz89: My suggestion is that NativeCurrency is an element of enum without TokenSymbol type. It already modify by @GopherJ in this pr. https://github.com/parallel-finance/parallel/pull/607/files#diff-9a08c969144a056406ced2d9dde2f410e0135639d066f770d14b702ebf909a80R113

runtime/parallel/src/lib.rs
yz89: We defined the `pub struct Adapter<AccountId>` in heiko, parallel and vanilla runtime. I mean if we can extract the logic into a pallet.

PR #591, opened 2021-09-15
Fix price decimal

MrPai:
Since decimal is a private type in `pallet_assets`, for now, decimal is defined in Runtime instead.


GopherJ (approved):

GopherJ:
@MrPai can you resolve conflicts, then I think we can merge

PR #607, opened 2021-09-17
Feat/remove orml tokens/currencies and CurrencyId

GopherJ:


GopherJ:
need to fix tests

GopherJ:
it's done now, should have been fixed

PR #620, opened 2021-09-19
Change big number to string literal

alannotnerd:


launch/config.ts
```diff
@@ -0,0 +1,73 @@
+import BN from 'bn.js';
+
+const config = {
```
alannotnerd, line 3: I guess it's better to use json file for other languages.

GopherJ: I'll test if string literal works, at the beginning I changed it to bignumber for testing if it's that causing errors

PR #630, opened 2021-09-23
fix CI temporary

yz89:


GopherJ:
why do we need to comment them? maybe we should just fix this problem

yz89:
> why do we need to comment them? maybe we should just fix this problem

It's difficult to fix now because we need to separate the benchmark to another crate or inherit the pallet_assets::Config.

GopherJ (approved):

Issue #376, opened 2021-08-14
Switch from xcm::v0 -> xcm::v1 & bump to polkadot-v0.9.10 (when it's released)


GopherJ:
so that we get `QueryResponse` support, blocked by: https://github.com/paritytech/polkadot/pull/3629, https://github.com/paritytech/cumulus/pull/571

GopherJ:
- [ ] bump to polkadot v0.9.9

GopherJ:
let's switch to xcm v1 first

PR #637, opened 2021-09-24
xcm v1

GopherJ:
close: #376 
close: #618
close: #641 

**TODO** before entering master:

- [x] build & publish parallelfinance/polkadot:v0.9.10 docker image
- [x] test cross chain transfer against the new docker image
- [x] fix loans benchmarks' logic
- [x] add loans benchmarks to runtimes
- [x] test make launch still works
- [x] switch to orml master when xcm v1 PR is merged
- [x] switch to polkadot-release-v0.9.10's final commit
- [x] update changelog
- [x] rename AssetIdentifier to CurrencyId after #639 

**TODO** after entering master:

-  [ ] parallel-js update, don't use AssetId, use CurrencyId instead
-  [ ] parallel-dapp should then use xcm v1 instead of xcm v0


runtime/heiko/src/lib.rs
```diff
@@ -399,10 +405,11 @@ parameter_types! {
     pub const MetadataDepositPerByte: Balance = deposit(0, 1);
 }
 
-impl pallet_assets::Config for Runtime {
+type AssetsInstance1 = pallet_assets::Instance1;
```
yz89, line 408: Is there a case that you need to use multiple assets instances?

pallets/loans/src/benchmarking.rs
```diff
@@ -36,28 +37,33 @@ const PENDING_MARKET_MOCK: Market = {
 
 const INITIAL_AMOUNT: u32 = 500_000_000;
 
-fn transfer_initial_balance<T: Config>(caller: T::AccountId)
-where
+fn transfer_initial_balance<
+    T: Config + pallet_assets::Config<AssetId = AssetIdentifier, Balance = Balance>,
```
yz89, line 41: Great work! What I don't understand is that Loans doesn't implement pallet_assets:Config if there is no inheritance. How does this work？

GopherJ: this is for limiting runtime used for benchmarks, `mock` and all runtimes implemented `pallet_assets::Config` so it worked

runtime/vanilla/Cargo.toml
```diff
@@ -20,71 +20,70 @@ smallvec          = '1.6.1'
...
+orml-oracle-rpc-runtime-api = { git = 'https://github.com/open-web3-stack/open-runtime-module-library.git', branch = 'sw/xcm-v1', default-features = false }
+orml-traits                 = { git = 'https://github.com/open-web3-stack/open-runtime-module-library.git', branch = 'sw/xcm-v1', default-features = false }
+orml-utilities              = { git = 'https://github.com/open-web3-stack/open-runtime-module-library.git', branch = 'sw/xcm-v1', default-features = false }
+orml-vesting                = { git = 'https://github.com/open-web3-stack/open-runtime-module-library.git', branch = 'sw/xcm-v1', default-features = false }
+orml-xcm                    = { git = 'https://github.com/open-web3-stack/open-runtime-module-library.git', branch = 'sw/xcm-v1', default-features = false }
+orml-xcm-support            = { git = 'https://github.com/open-web3-stack/open-runtime-module-library.git', branch = 'sw/xcm-v1', default-features = false }
+orml-xtokens                = { git = 'https://github.com/open-web3-stack/open-runtime-module-library.git', branch = 'sw/xcm-v1', default-features = false }
```
MrPai, line 86: need to wait for this to be merged into orml repo.

MrPai (commented):
All looks good, let's see when orml merge sw/xcm-v1.

runtime/vanilla/Cargo.toml
GopherJ: yes

GopherJ:
I cannot test cross chain yet, the current polkadot apps UI looks like:

image

Blocked by: https://github.com/polkadot-js/apps/issues/6199

GopherJ:
If I use MultiAsset v0 it'll work:

image

GopherJ:
encounterred `InvalidDest` while do upward transfer:

image

PR #640, opened 2021-09-24
convert spec to code and setup testing framework

drbh:
This draft PR converts https://www.notion.so/parallelfinance/pallet-crowdloans-a6a966a02e8347c793a154c7f64903b1 blueprint into a pallet we can test. 

Most functions are `unimplemented!();` and I was confused by `Box<Call>` so those args have been replaced by `Box<u8>` in order to successfully compile and test with `cargo test -p pallet-crowdloans`

This PR aims to implement most of the scaffolding we'll need for the pallet going forwards and is likely to change as we refine the spec.

### DIFF
--- a/pallets/crowdloans/src/lib.rs
+++ b/pallets/crowdloans/src/lib.rs
@@ -1,0 +1,250 @@
+// Copyright 2021 Parallel Finance Developer.
+// This file is part of Parallel Finance.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! # Crowdloans
+//!
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+extern crate alloc;
+
+mod crowdloan_structs;
+use crowdloan_structs::Vault;
+use crowdloan_structs::VaultPhase;
+
+use frame_support::{
+    pallet_prelude::*,
+    traits::{fungibles::Inspect, Get, IsType},
+    Blake2_128Concat, PalletId,
+};
+use frame_system::pallet_prelude::OriginFor;
+pub use pallet::*;
+use primitives::{currency::CurrencyId, Balance};
+
+#[cfg(test)]
+mod mock;
+#[cfg(test)]
+mod tests;
+
+pub type ParaId = u32;
+
+#[frame_support::pallet]
+pub mod pallet {
+    use super::*;
+    use frame_support::traits::tokens::fungibles;
+    use frame_system::ensure_root;
+    use primitives::AssetId;
+
+    use sp_arithmetic::traits::Zero;
+
+    #[pallet::config]
+    pub trait Config<I: 'static = ()>:
+        frame_system::Config + pallet_assets::Config<AssetId = AssetId, Balance = Balance>
+    {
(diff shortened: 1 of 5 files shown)


pallets/crowdloans/src/mock.rs
```diff
@@ -0,0 +1,294 @@
...
+        match asset {
+            CurrencyId::Native => <Balances as FungibleTransfer<AccountId>>::transfer(
+                source, dest, amount, keep_alive,
+            ),
+            CurrencyId::Asset(asset_id) => <Assets as Transfer<AccountId>>::transfer(
+                asset_id, source, dest, amount, keep_alive,
+            ),
+        }
+    }
+}
```
ETeissonniere, line 253: Shall we use the `pallet-currency-adapter` here?

pallets/crowdloans/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+                    contribution_strategy: ContributionStrategy::Placeholder(
+                        crowdloan,
+                        currency_shares,
+                        0,
+                    ),
+                    claim_strategy: ClaimStrategy::Placeholder(crowdloan),
+                    contributed: 0,
+                }))
+            })
+        }
```
ETeissonniere, line 167: We should make sure no vault for the same parachain already exists

drbh: @ETeissonniere i'm not sure I understand how the check above https://github.com/parallel-finance/parallel/blob/14ebaade49171edf537329c432413f751aa7377b/pallets/crowdloans/src/lib.rs#L144-L148 is insufficient? Where else should be checked before mutating? Thanks!

pallets/crowdloans/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+            let currency_shares_issuance = T::CrowdloanCurrency::total_issuance(currency_shares);
+
+            // make sure both project_shares and currency_shares are new assets
+            ensure!(
+                project_shares_issuance == Zero::zero() && currency_shares_issuance == Zero::zero(),
+                Error::<T>::SharesNotNew
+            );
+
+            // make sure no similar vault already exists as identified by crowdloan
+            ensure!(
```
ashi20192020ayush, line 145: We can move this check under `try_mutate` to avoid one extra read call.

pallets/crowdloans/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+
+        /// Contribute `amount` to the vault of `crowdloan` and receive some
+        /// shares from it
+
+        #[pallet::weight(10_000)]
+        #[allow(unused)]
+        pub fn contribute(
+            origin: OriginFor<T>,
+            crowdloan: ParaId,
+            amount: Balance,
```
ashi20192020ayush, line 180:
```suggestion
            amount: BalanceOf<T>,
```

pallets/crowdloans/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        /// If a `crowdloan` failed, claim back your share of the assets you
+        /// contributed
+        #[pallet::weight(10_000)]
+        #[allow(unused)]
+        pub fn claim_refund(
+            origin: OriginFor<T>,
+            crowdloan: ParaId,
+            amount: Balance,
```
ashi20192020ayush, line 229:
```suggestion
            amount: BalanceOf<T>,
```

pallets/crowdloans/src/mock.rs
```diff
@@ -0,0 +1,294 @@
...
+    type Freezer = ();
+    type Extra = ();
+    type WeightInfo = ();
+}
+
+parameter_types! {
+    pub const CrowdloanPalletId: PalletId = PalletId(*b"par/cwlp");
+}
+
+pub struct Adapter<AccountId> {
```
ashi20192020ayush, line 154: Can we use CurrencyAdapter like this https://github.com/parallel-finance/parallel/blob/master/pallets/amm/src/mock.rs#L183

ashi20192020ayush (changes requested):

PR #673, opened 2021-10-07
contains the weight of accrue_interest to the block

### LINKED ISSUE #671: Contains the weights of accrue_interest to the block

yz89:
close #671


GopherJ:
other than those comments it should look good

PR #688, opened 2021-10-12
Automatically prepare xcm fees for settlement

alannotnerd:
Send fixed transaction compensation to relaychain account while do settlement.


GopherJ:
@alannotnerd @yz89

PR #720, opened 2021-10-16
Add staking cap

alannotnerd:
Add limitation for liquid staking. While `StakingPool` reachs the bound, `stake` will fail and throw `ExceededStakingPool` error.
But it will still increase, because we don't check the capacity during recording rewards.


alannotnerd:
> there are a few changes to be made before merging:
> 
> 1. we didn't use force_* prefix before for governance call, add this prefix is normally only for sudo, I think we should remove this prefix and handle it later for all if we want(I think we don't need)
> 2. launch script should be adjusted
> 3. xcm fees compensation's default value shouldn't be zero

I've removed the prefix and adjust compensation as well as poll cap in launch script.

Issue #725, opened 2021-10-16
Burn fees in each transact call


GopherJ:

GopherJ:
we can put burn inside `ump_transact` and `Withdraw` should also use `xcm_fees_compensation` value

GopherJ:
we should also explain why do we burn

PR #733, opened 2021-10-18
burn fees in ump_transact instead

0x8f701:
close: #725 

TODO:

- [x] fix tests


GopherJ:
let's merge it so that we can test this afternoon

Issue #787, opened 2021-10-22
Add error handler or callback for liquid staking's ump transact


GopherJ:
- bond => burn after success, ledger.total_stake - bonded_amount
- bond_extra => burn after success, ledger.total_stake - bonded_amount
- withdraw_unbonded => mint after success
- unbonded => ledger.total_unstake - unbonded_amount
- rebond => ledger.total_stake_amount - rebond_amount

PR #801, opened 2021-10-24
finish transact notify feature

GopherJ:
close: #787 

TODO:

- [ ] test work as expected
- [ ] fix callback's call's weight
- [ ] fix unit tests
- [ ] enrich with error index if failed

Blocker:

1. https://github.com/paritytech/polkadot/issues/4139


GopherJ:
ok this is because that the function signature must be `origin, query_id, response`

PR #842, opened 2021-11-01
refactor price pallet

yz89:


GopherJ (approved):

GopherJ:
@yz89 can you use branch in parallel-finance~ that makes co-working easier, like `yz/refactor-prices`

PR #867, opened 2021-11-09
fix launch scripts

GopherJ:


GopherJ:
TODO:

some clients should also use 9948

PR #872, opened 2021-11-10
Extract ensure_liquidity method

rjman-ljm:
chore

### DIFF
--- a/pallets/loans/src/ptoken.rs
+++ b/pallets/loans/src/ptoken.rs
@@ -206,12 +206,10 @@ where
         let market = Self::ensure_active_market(underlying_id)?;
         let collateral_value = Self::collateral_asset_value(who, underlying_id)?;
 
-        // liquidity of all assets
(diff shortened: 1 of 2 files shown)


pallets/loans/src/ptoken.rs
```diff
@@ -206,12 +206,10 @@ where
         let market = Self::ensure_active_market(underlying_id)?;
         let collateral_value = Self::collateral_asset_value(who, underlying_id)?;
 
-        // liquidity of all assets
-        let (liquidity, _) = Self::get_account_liquidity(who)?;
-
-        if liquidity >= collateral_value {
+        if let Ok(()) = Self::ensure_liquidity(who, collateral_value) {
```
yz89, line 209: There is no need to change it because liquidity is used down here

rjman-ljm: Okay, I will change it back

yz89 (approved):

PR #894, opened 2021-11-18
enable lto

GopherJ:
close: #891


GopherJ:
this cannot be built via CI as we dont have enough resouces

Issue #897, opened 2021-11-18
HKO, PARA's decimal provider


GopherJ:
if we want to lend HKO/PARA in money market they should have a decimal. The current decimal provider uses Assets' metadata but native tokens are not an asset yet.

- if we want to lend HKO/PARA, then we will need to register a new asset HKO/PARA => should be ok, we have CurrencyAdapter, but how to prevent transfer via Assets because it makes nosense

yz89:
Money market won't lend the ptoken ever. Instead, ptokens are derivatives of other deposit currencies.

GopherJ:
> Money market won't lend the ptoken ever. Instead, ptokens are derivatives of other deposit currencies.

yes, then it's cyclic if we do so

GopherJ:
Looks like native tokens' decimal should be managed by Assets module so that we dont need to create an asset

PR #900, opened 2021-11-18
Feat/distinguish decimal between native token and assets

rjman-ljm:
## Issues

Close
- #897

### DIFF
--- a/pallets/loans/src/mock.rs
+++ b/pallets/loans/src/mock.rs
@@ -139,8 +139,12 @@ impl ExchangeRateProvider for LiquidStakingExchangeRateProvider {
     }
 }
 
-pub struct Decimal;
-impl DecimalProvider for Decimal {
+pub struct Currencies;
+impl CurrencyProvider<CurrencyId, CurrencyType> for Currencies {
+    fn get_currency_type(_currency_id: &CurrencyId) -> CurrencyType {
+        CurrencyType::Unknown
+    }
+
     fn get_decimal(asset_id: &CurrencyId) -> Option<u8> {
         match *asset_id {
             KSM | XKSM => Some(12),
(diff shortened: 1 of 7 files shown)


primitives/src/lib.rs
```diff
@@ -101,7 +101,17 @@ pub trait PriceFeeder {
     fn get_price(asset_id: &CurrencyId) -> Option<PriceDetail>;
 }
 
-pub trait DecimalProvider {
+pub enum CurrencyType {
```
yz89, line 104: There is no need to add this type.

runtime/heiko/src/lib.rs
```diff
@@ -973,14 +975,32 @@ impl DataProviderExtended<CurrencyId, TimeStampedPrice> for AggregatedDataProvid
     }
 }
 
-pub struct Decimal;
-impl DecimalProvider for Decimal {
+pub struct Currencies;
+impl CurrencyProvider<CurrencyId, CurrencyType> for Currencies {
+    fn get_currency_type(asset_id: &CurrencyId) -> CurrencyType {
```
yz89, line 980: We just need to distinguish whether the token is a native token or from pallet_asset.

pallets/loans/src/mock.rs
```diff
@@ -173,7 +177,7 @@ impl pallet_prices::Config for Test {
     type FeederOrigin = EnsureRoot<AccountId>;
     type LiquidStakingExchangeRateProvider = LiquidStaking;
     type LiquidStakingCurrenciesProvider = LiquidStaking;
-    type Decimal = Decimal;
+    type CurrencyProvider = Currencies;
```
yz89, line 180: No need to modify the Config.

runtime/heiko/src/lib.rs
rjman-ljm: okay, got it.

GopherJ (approved):

yz89 (approved):

PR #902, opened 2021-11-19
Feat/a usd

GopherJ:


GopherJ:
HKO cross chain from vanilla-dev -> karura-dev is working:

GopherJ:
HKO transfer from karura-dev -> vanilla-dev fails:

image

PR #913, opened 2021-11-23
Emergency shutdown

ashi20192020ayush:


runtime/heiko/src/lib.rs
```diff
@@ -246,19 +246,29 @@ impl Contains<Call> for BaseCallFilter {
...
         // // LiquidStaking
         // Call::LiquidStaking(_) |
         // Call::NomineeElection(_) |
 
         // // Membership
-        // Call::ValidatorFeedersMembership(_)
+        // Call::ValidatorFeedersMembership(_) |
+        // Call::OracleMembership(_)
+    }
+}
```
ETeissonniere, line 260: We need to define a new filter here. Not patch the existing one.

runtime/heiko/src/lib.rs
```diff
@@ -1310,6 +1320,12 @@ impl pallet_currency_adapter::Config for Runtime {
 //     type WeightInfo = pallet_liquidity_mining::weights::SubstrateWeight<Runtime>;
 // }
 
+impl pallet_emergency_shutdown::Config for Runtime {
+    type Event = Event;
+    type Whitelist = WhiteListFilter;
```
ETeissonniere, line 1325: This should be the new filter we define previously

ETeissonniere (changes requested):

GopherJ:
I'm still a bit confused on this. What if we only want to shutdown crowdloan pallet's certain calls + loans pallet's certain calls?

ETeissonniere:
@GopherJ that would not be possible. The goal of the emergency shutdown pallet is to shut down all non-critical operations in the event of a major issue. If we wanted to resume operations for one pallet or the other this would have to go through either:
1. Disabling the emergency shut down state all together
2. Issuing a runtime upgrade to reactivate some of the pallets we think are safe to reopen

GopherJ:
> @GopherJ that would not be possible. The goal of the emergency shutdown pallet is to shut down all non-critical operations in the event of a major issue. If we wanted to resume operations for one pallet or the other this would have to go through either:
> 
>     1. Disabling the emergency shut down state all together
> 
>     2. Issuing a runtime upgrade to reactivate some of the pallets we think are safe to reopen

ok got it

runtime/heiko/src/lib.rs
```diff
@@ -1454,6 +1502,10 @@ construct_runtime!(
 
         // Bridge
         Bridge: pallet_bridge::{Pallet, Call, Storage, Event<T>} = 90,
+
+        // Emergency Shutdown
+        EmergencyShutdown: pallet_emergency_shutdown::{Pallet, Call, Event<T>} = 84,
```
GopherJ, line 1507: 84 shouldn't be put after 90

runtime/parallel/src/lib.rs
```diff
@@ -1376,6 +1383,47 @@ impl pallet_currency_adapter::Config for Runtime {
...
+            Call::Utility(_) |
+            Call::Multisig(_) |
+            Call::Proxy(_) |
+            // 3rd Party
+            Call::Vesting(_) |
+            // Membership
+            Call::GeneralCouncilMembership(_) |
+            Call::TechnicalCommitteeMembership(_) |
+			// Emergency Shutdown pallet
+			Call::EmergencyShutdown(_)
```
GopherJ, line 1416: this should be aligned

runtime/vanilla/src/lib.rs
```diff
@@ -1406,6 +1413,47 @@ impl pallet_liquidity_mining::Config for Runtime {
...
+            Call::Utility(_) |
+            Call::Multisig(_) |
+            Call::Proxy(_) |
+            // 3rd Party
+            Call::Vesting(_) |
+            // Membership
+            Call::GeneralCouncilMembership(_) |
+            Call::TechnicalCommitteeMembership(_) |
+			// Emergency Shutdown pallet
+			Call::EmergencyShutdown(_)
```
GopherJ, line 1446: should be aligned

runtime/heiko/src/lib.rs
```diff
@@ -1381,6 +1388,47 @@ impl pallet_currency_adapter::Config for Runtime {
...
+            Call::Utility(_) |
+            Call::Multisig(_) |
+            Call::Proxy(_) |
+            // 3rd Party
+            Call::Vesting(_) |
+            // Membership
+            Call::GeneralCouncilMembership(_) |
+            Call::TechnicalCommitteeMembership(_) |
+			// Emergency Shutdown pallet
+			Call::EmergencyShutdown(_)
```
GopherJ, line 1421: should be aligned

runtime/parallel/src/lib.rs
```diff
@@ -1448,6 +1496,9 @@ construct_runtime!(
         // LiquidityMining: pallet_liquidity_mining::{Pallet, Call, Storage, Event<T>} = 83,
         // Bridge
         Bridge: pallet_bridge::{Pallet, Call, Storage, Event<T>} = 90,
+
+        // Emergency Shutdown
+        EmergencyShutdown: pallet_emergency_shutdown::{Pallet, Call, Event<T>} = 84,
```
GopherJ, line 1501: this index isn't correct

runtime/vanilla/src/lib.rs
```diff
@@ -1478,6 +1526,9 @@ construct_runtime!(
 
         // Bridge
         Bridge: pallet_bridge::{Pallet, Call, Storage, Event<T>} = 90,
+
+        // Emergency Shutdown
+        EmergencyShutdown: pallet_emergency_shutdown::{Pallet, Call, Event<T>} = 84,
```
GopherJ, line 1531: this index looks incorrect

runtime/heiko/src/lib.rs
```diff
@@ -256,9 +256,16 @@ impl Contains<Call> for BaseCallFilter {
     }
 }
 
+pub enum CallFilterRouter {}
```
GopherJ, line 259: empty struct is better

pallets/emergency-shutdown/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+            // Emit an event.
+            Self::deposit_event(Event::ToggledShutdownFlag(updated_flag));
+            Ok(())
+        }
+    }
+}
+
+impl<T: Config> Contains<T::Call> for Pallet<T> {
+    fn contains(call: &T::Call) -> bool {
+        if Self::is_shut_down_flag() {
```
GopherJ, line 83: can be simplified by:

```rust
!Self::is_shut_down_flag() ||  T::Whitelist::contains(call)
```

ETeissonniere (approved):

PR #917, opened 2021-11-24
give HKO/PARA as rewards while receiving 10DOT 1KSM

GopherJ:
TODO:

- [x]  test using make launch
- [x] fees should cover minted PARA

### DIFF
--- a/primitives/src/currency.rs
+++ b/primitives/src/currency.rs
@@ -12,9 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use frame_support::traits::{
-    fungibles::{Inspect, Mutate, Transfer},
-    Get,
+use frame_support::{
+    log,
+    traits::{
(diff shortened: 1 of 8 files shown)


GopherJ:
tested working:
image

primitives/src/currency.rs
```diff
@@ -76,6 +103,34 @@ impl<
...
+
+                    if !gift_amount.is_zero() && reducible_balance >= gift_amount {
+                        if let Err(e) = MultiCurrency::transfer(
+                            native_currency_id,
+                            &gift_account,
+                            &who,
+                            gift_amount,
+                            false,
+                        ) {
+                            return Err(XcmError::FailedToTransactAsset(e.into()));
```
GopherJ, line 129: maybe I shouldn't let it return because this can be dangerous...

GopherJ:
double direction cross chain transfer consumed 0.0001 KSM ~ 0.2 HKO

PR #922, opened 2021-11-25
add liquidstaking log

### LINKED ISSUE #919: add logs for liquid staking

MrPai:
currently, no need for more logs, because the method will return error info which is enough now
close #919 #930

### DIFF
--- a/pallets/liquid-staking/src/lib.rs
+++ b/pallets/liquid-staking/src/lib.rs
@@ -353,7 +353,7 @@ pub mod pallet {
                 if let Err(err) =
                     T::Assets::transfer(staking_currency, &account_id, who, *amount, false)
                 {
-                    log::error!("Transfer failed {:?}", err);
+                    log::error!(target: "liquidstaking::on_idle", "Transfer failed {:?}", err);
                     // break if we cannot afford this
                     break;
                 }
--- a/pallets/liquid-staking/src/lib.rs
+++ b/pallets/liquid-staking/src/lib.rs
@@ -41,7 +41,7 @@ pub use pallet::*;
 pub mod pallet {
     use frame_support::{
         dispatch::{DispatchResult, DispatchResultWithPostInfo},
-        ensure,
+        ensure, log,
         pallet_prelude::*,
         require_transactional,
         traits::{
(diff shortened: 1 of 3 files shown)


pallets/liquid-staking/src/lib.rs
```diff
@@ -337,12 +337,23 @@ pub mod pallet {
...
+                );
                 if free_balance < *amount {
                     return remaining_weight;
                 }
 
-                if T::Assets::transfer(staking_currency, &account_id, who, *amount, false).is_err()
+                if let Err(err) =
+                    T::Assets::transfer(staking_currency, &account_id, who, *amount, false)
                 {
+                    log::error!("Transfer failed {:?}", err);
```
GopherJ, line 356: shall we have also a target?

pallets/liquid-staking/src/lib.rs
```diff
@@ -1040,6 +1051,14 @@ pub mod pallet {
             let account_id = Self::account_id();
             let asset: MultiAsset = (MultiLocation::here(), fees).into();
 
+            log::trace!(
+                target: "liquidstaking::ump_transact",
+                "call: {:?}, asset: {:?}, xcm_weight: {:?}",
```
GopherJ, line 1056: this asset information will not be very useful because it's only relaychain's native currency

GopherJ, line 1056: have you tested if `call` is human readable because if it's not then it wont be very useful

MrPai: It will be helpful too if it's a raw bytes vec, remember last time I compare two bytes vec between parachain xcm message and relaychain xcm.

MrPai: I think it conclude a fee number, which may be good for debug, as you know, this is trace level log.

pallets/liquid-staking/src/lib.rs
MrPai: ok

alannotnerd (approved):

pallets/liquid-staking/src/lib.rs
GopherJ: ok we can do this first I think

PR #953, opened 2021-11-30
Support external token for bridge and add handling fee

rjman-ljm:
Close
  - #977


rjman-ljm:
I'm working on `extends assets and membership for bridge` now, maybe review it later

pallets/bridge/src/lib.rs
```diff
@@ -278,47 +286,81 @@ pub mod pallet {
...
+        #[transactional]
+        pub fn set_currency_fee(
+            origin: OriginFor<T>,
+            currency_id: CurrencyId,
+            new_fee: BalanceOf<T>,
+        ) -> DispatchResult {
+            Self::ensure_admin(origin)?;
+            Self::ensure_currency_registered(currency_id)?;
+
+            let asset_id = AssetIds::<T>::get(currency_id);
```
yz89, line 345: Why not use currency_id? It confuses that getting an asset_id by currency_id and modify the Currencies by asset_id.

pallets/bridge/src/lib.rs
```diff
@@ -333,11 +375,28 @@ pub mod pallet {
             Self::ensure_currency_registered(currency_id)?;
 
             let asset_id = AssetIds::<T>::get(currency_id);
-            T::Assets::transfer(asset_id, &who, &Self::account_id(), amount, false)?;
+            let Currency { external, fee, .. } = Currencies::<T>::get(asset_id);
+            let total_amount = amount.checked_add(fee).ok_or(ArithmeticError::Overflow)?;
+            if external {
+                T::Assets::burn_from(asset_id, &who, total_amount)?;
+            } else {
+                T::Assets::transfer(asset_id, &who, &Self::account_id(), total_amount, false)?;
```
yz89, line 383: Will the relayer get the fee information from Currencies storage and reserve the fee itself?

yz89, line 383: Will the relayer get the fee information from Currencies storage and reserve the fee itself?

pallets/bridge/src/lib.rs
rjman-ljm: *Reason*
the struct of currency is complex，it's not suitable as a key of maapping

*Solutions*
1. currency_id => asset_id => currency (Just like above👆)
2. currency_id => currency (need a extra storage)

pallets/bridge/src/lib.rs
rjman-ljm: No, relayer only processes the transaction itself and does not make changes to the transaction information（like the fee）

rjman-ljm: Fee will be reserved by `teleported`, if we want to get these reserved fees, we can sendAndSign `materialize`, just like the cross-chain extrinsics do

rjman-ljm: No, relayer only processes the transaction itself and does not make changes to the transaction information. Fees will be reserved by `teleport`, if we want to get these reserved fee, we can sendAndSign the `materialize`, just like the cross-chain extrinsics do

pallets/bridge/src/lib.rs
rjman-ljm: This did cause a bit confusion, maybe I should add another storage for `currency_id => currency`?

pallets/bridge/src/lib.rs
rjman-ljm: Same as above

yz89 (approved):

PR #958, opened 2021-11-30
use batchAll to send also remark for more info

GopherJ:


yz89:
vault id # blocknumber # user account

PR #975, opened 2021-12-02
make crowdloan duration shorter

GopherJ:
close: #969

### DIFF
--- a/launch/config.ts
+++ b/launch/config.ts
@@ -181,8 +181,8 @@ const config = {
       image: 'parallelfinance/polkadot-collator:v0.9.12',
       chain: 'shell',
       ctokenId: 4000,
-      cap: '10000000000000000',
-      duration: 202800,
+      cap: '100000000000000',
+      duration: 28800,
       leaseStart: 0,
       leaseEnd: 7
     },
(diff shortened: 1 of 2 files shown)


launch/config.ts
```diff
@@ -180,7 +181,8 @@ const config = {
       image: 'parallelfinance/polkadot-collator:v0.9.12',
       chain: 'shell',
       ctokenId: 4000,
-      cap: '100000000000000000',
+      cap: '100000000000000',
```
i5possible, line 184: Why this cap is changed?

launch/index.ts
```diff
@@ -189,17 +189,17 @@ async function relay() {
 
   console.log('Start new auction.')
   const call = []
-  call.push(api.tx.sudo.sudo(api.tx.auctions.newAuction(1000000, config.leaseIndex)))
+  call.push(api.tx.sudo.sudo(api.tx.auctions.newAuction(config.auctionDuration, config.leaseIndex)))
```
i5possible, line 192: Change from 1000000 to 201600? Is that means 56 minutes？

launch/config.ts
GopherJ: @i5possible because we would like to make it easier for testing the edge case

i5possible: So we are not goint to deploy it to prod and just for testing usage?

GopherJ: this script is only made for testing purpose because on prod we cannot control relaychain at all. It's up to parity

Issue #968, opened 2021-12-01
Support contribution before auction


GopherJ:

GopherJ:
we can have a pending contributions pool. This pool can gather users' contributions and then contribute on relay chain once a real contribution happens

GopherJ:
need a flag to distinguish if a vault isn't ready to contribute => pending

GopherJ:
Vault {
...
pending: number
}

VaultPhase => ... | Pending

open(paraId)

PR #980, opened 2021-12-04
Feat/early contribution

GopherJ:
close: #968


GopherJ:
do you have better idea of `early contribution` @ETeissonniere ? now we use pending to record what haven't been contributed on relaychain at the stage `Pending`

After calling `open` it enters the phase `Contributing` then the next contribute call will bring all all `pending` to relaychain and clean `pending`

pallets/crowdloans/src/lib.rs
```diff
@@ -268,6 +270,28 @@ pub mod pallet {
...
+            Vaults::<T>::try_mutate(&crowdloan, |vault| -> Result<_, DispatchError> {
+                let mut vault = vault.as_mut().ok_or(Error::<T>::VaultDoesNotExist)?;
+
+                ensure!(
+                    vault.phase == VaultPhase::Pending,
+                    Error::<T>::IncorrectVaultPhase
+                );
+
+                vault.phase = VaultPhase::Contributing;
+
```
ETeissonniere, line 288: Shouldn't we call `do_contribute` here and update the state accordingly?

ETeissonniere:
@GopherJ I like the way it is currently, shall we just auto contribute what is pending during the `open` call?

GopherJ:
> 

Yes I think it's a good idea!

PR #981, opened 2021-12-04
offchain project token distribution

drbh:


drbh:
@GopherJ

pallets/crowdloans/src/lib.rs
```diff
@@ -531,6 +554,78 @@ pub mod pallet {
...
+                for (ctoken, percent_of_ctokens) in accounts_ownership.iter() {
+                    // get the total number of ptokens to be distributed
+                    let ptoken = Self::get_ptoken_from_ctoken(ctoken);
+                    let current_distribution_total = Self::get_distribution_amount(ptoken);
+
+                    // the user should recieve this portion of ptokens
+                    let amount = current_distribution_total * percent_of_ctokens;
+
+                    // transfer tokens to user from this account
+                    T::Assets::transfer(ptoken, &Self::account_id(), &account_id, amount, true)
```
GopherJ, line 580: we should mint project tokens directly on our chain. Because the tokens are on another chain (recorded on parachain account - 2012/2085), so we can mint it directly from our chain.

GopherJ: This needs to be tested and validated, I'll test it

pallets/crowdloans/src/lib.rs
```diff
@@ -531,6 +554,78 @@ pub mod pallet {
...
+                        account_id.clone(),
+                        amount,
+                    ));
+                }
+            }
+            Ok(())
+        }
+
+        // TODO: how do we determine the ptoken for each ctoken
+        fn get_ptoken_from_ctoken(_ctoken: &u32) -> u32 {
```
GopherJ, line 595: we can have a mapping to manage the relationship:

ctoken -> (ParaId, Auctions BatchIndex, project tokenId)

GopherJ: but then when we create a vault we will need to have already created two tokens:

- ctoken
- ptoken

pallets/crowdloans/src/lib.rs
```diff
@@ -531,6 +554,78 @@ pub mod pallet {
...
+        }
+
+        // TODO: how do we determine the balance of tokens to distribute this iteration
+        fn get_distribution_amount(_ptoken: u32) -> u128 {
+            10_000
+        }
+
+        pub fn ctoken_ownership_snapshot() -> BTreeMap<T::AccountId, BTreeMap<u32, u128>> {
+            // iterate over all accounts on chain and build ownership map for each
+            frame_system::Account::<T>::iter()
```
GopherJ, line 606: this is two much... I still prefer not using OCW because we had bad memory about OCW in liquidation pallet, it affected a lot the block production time

GopherJ:
I still think we shouldn't manage this on-chain, ping @yz89

yz89:
We need to talk more about this design.

pallets/crowdloans/src/lib.rs
```diff
@@ -531,6 +554,78 @@ pub mod pallet {
...
+
+                            // vault.ctoken, who, percent_of_payout
+                            (vault.ctoken, percent_of_payout)
+                        })
+                        .collect(); // we collect in to a variable for readability
+                    (who, ctoken_ownership_map)
+                })
+                .collect()
+        }
+
```
ETeissonniere, line 628: I am wondering if there could be a race condition here where if we take more than one block, or 12 seconds, to execute this function we may have a case where some of the balances change while this function is still running

GopherJ:
let's pause this PR @drbh util we got more ideas on it

Issue #962, opened 2021-12-01
the same ParaId can be used multiple times OR by multiple projects


GopherJ:

GopherJ:
New storages:

- LastIndexes => Map { ParaId -> u32 }
- CTokens => Map { ctokenId => ParaId }

Existing calls:

- create_vault
   1. check if LastIndexes has this ParaId, if yes and the correspond vault is expired / failed, then increment LastIndexes and create new vault
   2. put new ctoken into CTokens map

- contribute
  1. get last_index and check if vault is contributing, if yes then allow to contribute

- close
  1. get last_index and check if vault is contributing, if yes then close it
 
- reopen
   1. get last_index and check if vault is closed, if yes then reopen it

- auction_failed
   1. get last_index and check if vault is closed, if yes then fail it

- slot_expired
   1. get last_index and check if vault is closed, if yes then expire it

GopherJ:
it seems ctoken shouldn't be released to users util the parathread won a slot

GopherJ:
how many types of ctokens will we have for the same parathread?

PR #997, opened 2021-12-07
Feat/multiple vaults

GopherJ:
close: #962


ETeissonniere (approved):
lgtm. Just wondering why we need to identify the vault per cToken instead of ParaId. We could simply overwrite the previous vault of it is failed. Nonetheless, it works as it is so approved it anyways.

GopherJ:
@ETeissonniere because people can be slow to claim back their tokens actually

pallets/crowdloans/src/weights.rs
```diff
@@ -65,62 +65,62 @@ pub trait WeightInfo {
 pub struct SubstrateWeight<T>(PhantomData<T>);
 impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
     fn create_vault() -> Weight {
-        (39_000_000 as Weight)
```
i5possible, line 68: Maybe a stupid question: How is the weight is calculated or set?

PR #1061, opened 2021-12-16
check all balance inputs gt zero

drbh:
closes https://github.com/parallel-finance/parallel/issues/1007

### DIFF
--- a/pallets/crowdloans/src/lib.rs
+++ b/pallets/crowdloans/src/lib.rs
@@ -258,6 +258,8 @@ pub mod pallet {
         ) -> DispatchResult {
             T::CreateVaultOrigin::ensure_origin(origin)?;
 
+            ensure!(cap > 0, Error::<T>::InsufficientBalance);
+
             let ctoken_issuance = T::Assets::total_issuance(ctoken);
             ensure!(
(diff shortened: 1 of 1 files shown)


pallets/crowdloans/src/lib.rs
```diff
@@ -258,6 +258,8 @@ pub mod pallet {
         ) -> DispatchResult {
             T::CreateVaultOrigin::ensure_origin(origin)?;
 
+            ensure!(cap > 0, Error::<T>::InsufficientBalance);
```
GopherJ, line 261: we may need to give more explicit error. and `cap > 0` can be replaced by `!cap.is_zero()`

pallets/crowdloans/src/lib.rs
```diff
@@ -539,6 +542,8 @@ pub mod pallet {
         ) -> DispatchResult {
             let who = ensure_signed(origin)?;
 
+            ensure!(amount > 0, Error::<T>::InsufficientBalance);
```
GopherJ, line 545: The error isn't good

GopherJ: maybe just `InvalidParams`

GopherJ (approved):

PR #1066, opened 2021-12-19
Fix audit findings

rustaceanstayhigh:
Check these tasks that's in progress. [In Progress](https://www.notion.so/parallelfinance/6ae7ee23a88d4abc8ce8f0be4d88c143?v=11eb497b9b104d999e97b520ad570435)

### DIFF
--- a/pallets/crowdloans/src/lib.rs
+++ b/pallets/crowdloans/src/lib.rs
@@ -212,6 +212,12 @@ pub mod pallet {
         ExceededMaxVrfs,
         /// Pending contribution must be killed before entering `Contributing` vault phase
         PendingContributionNotKilled,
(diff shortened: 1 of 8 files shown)


pallets/crowdloans/src/lib.rs
```diff
@@ -635,6 +646,31 @@ pub mod pallet {
             xcm_weight_misc: XcmWeightMisc<Weight>,
         ) -> DispatchResultWithPostInfo {
             T::UpdateOrigin::ensure_origin(origin)?;
+
+            let XcmWeightMisc {
```
GopherJ, line 650: This may shouldn't be in Call, can you make it a member function for `XcmWeightMisc`

PR #1124, opened 2022-01-03
Pallet crowdloans xcm simulator

### LINKED ISSUE #1096: migrate pallet-crowdloans' unit tests to xcm-simulator

MrPai:
close #1096 
Xcm `ReportError` can not test in xcm simulator because it needs to be processed at another block.
In simulator, it seems like parachain side will receive dmp message however can not process it.

Would like to merge this PR because it shows how to use xcmconfig on relaychain side, may refer it future.

xcm

### DIFF
--- a/pallets/crowdloans/src/mock.rs
+++ b/pallets/crowdloans/src/mock.rs
@@ -9,7 +9,7 @@ use frame_support::{
 use frame_system::{EnsureOneOf, EnsureRoot, EnsureSignedBy};
 use orml_xcm_support::IsNativeConcrete;
 use pallet_xcm::XcmPassthrough;
-use polkadot_parachain::primitives::Sibling;
+use polkadot_parachain::primitives::{HeadData, Sibling, ValidationCode};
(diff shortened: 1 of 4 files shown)


pallets/crowdloans/src/mock.rs
```diff
@@ -516,7 +566,7 @@ pub fn para_ext(para_id: u32) -> sp_io::TestExternalities {
         System::set_block_number(1);
         Assets::force_create(Origin::root(), DOT, Id(ALICE), true, 1).unwrap();
         Assets::force_create(Origin::root(), XDOT, Id(ALICE), true, 1).unwrap();
-        Assets::mint(Origin::signed(ALICE), DOT, Id(ALICE), 100 * DOT_DECIMAL).unwrap();
```
GopherJ, line 519: I think we have helper function `dot()`

GopherJ: maybe we can change this later

Issue #1086, opened 2021-12-24
Switch from cDOT-project to cDOT-lease


GopherJ:
## Solutions

1. reserve token while minting cDOT to user, users need to come to us to unreserve (only work for cDOT-project it seems)
2. dont mint cDOT after contribution, issue cDOT after winning (via on_idle + childtrie)
3. dont mint cDOT after contribution, issue cDOT after winning (via offchain-client)

GopherJ:
now we decide to insist using cDOT-project

GopherJ:
## Scope

1. calculate cDOT rewards
2. create cDOT asset for batch 1
3. mint cDOT-batch1 for contributors
4. use StorageMigration to fix vault

GopherJ:
## cDOT lease implementation guide

### Storage

1. remove `BatchIndexes` 
2. add `OngoingCrowdloans`
3. change CTokensRegistry`'s key to be `(leaseStart, leaseEnd)`
4. add leaseStart, leaseEnd parameter for `claim_refund`

Issue #1069, opened 2021-12-20
Missing LeaseStart, LeaseEnd fields in vault


GopherJ:

GopherJ:
This may be useful for frontend

PR #1138, opened 2022-01-06
cdot-lease

GopherJ:

close: #1086 
close: #1069


pallets/crowdloans/src/lib.rs
```diff
@@ -199,7 +199,7 @@ pub mod pallet {
         AllMigrated(ParaId),
         /// Partially contributions migrated
         /// [para_id, non_migrated_count]
-        PartiallyMigrated(ParaId, VaultId),
+        PartiallyMigrated(ParaId, u32),
```
ashi20192020ayush, line 202: Can we use `TrieIndex` instead of u32

GopherJ: @ashi20192020ayush it's for `non_migrated_count` so better be u32

PR #1160, opened 2022-01-11
Add batch1 valuts

### LINKED ISSUE #1135: add vaults for batch 1 winning projects (Acala, moonbeam...)
based on: https://github.com/parallel-finance/parallel/issues/1086#issuecomment-1004561593

MrPai:
close #1135 

- [x] Need to confirm ctoken asset Id
- [x] Need to use actually contributed amount rather than raised amount
refer to https://www.notion.so/parallelfinance/Partners-Reward-Vesting-eaa15f82161b44e984a17f5ea29d8698 (finalized)

**NOTE**
the contributed amount includes the amount that users send to our multi-sig, however, we have not sent it to relaychain crowdloan.


pallets/crowdloans/src/migrations.rs
```diff
@@ -0,0 +1,111 @@
...
+//! # Add vaults for batch 1 winning projects
+
+use super::*;
+use frame_support::traits::Get;
+use primitives::ParaId;
+use sp_runtime::traits::Zero;
+use sp_std::{vec, vec::Vec};
+use types::*;
+/// Add vaults for batch 1 winning projects
+pub fn migrate<T: Config>() -> frame_support::weights::Weight {
```
GopherJ, line 24: Since we have storage migration, maybe we should have a version management

GopherJ: we can call `migrations::v?::migrate` in runtime

MrPai: Good point, do you think if we need storage like this:
https://github.com/paritytech/substrate/blob/961666a2cb/frame/staking/src/pallet/mod.rs#L462
and an Enum like this:
https://github.com/paritytech/substrate/blob/961666a2cb/frame/staking/src/lib.rs#L765-L774

GopherJ: Yes I think so

GopherJ:
LGTM, I just dont know if we need a storage to record current version.

This can be merged into master when we finish the cdot mint script and runtime upgrade tests. we can then start to deploy to heiko

GopherJ (approved):

Issue #1164, opened 2022-01-11
AMM locked minimum liquidity


GopherJ:
https://github.com/Uniswap/v2-core/blob/master/contracts/UniswapV2Pair.sol#L120-L124

PR #1179, opened 2022-01-14
adds minimum liquidity

drbh:
resolves https://github.com/parallel-finance/parallel/issues/1164

### DIFF
--- a/pallets/amm/src/benchmarking.rs
+++ b/pallets/amm/src/benchmarking.rs
@@ -21,6 +21,7 @@ const BASE_ASSET: CurrencyId = XDOT;
 const QUOTE_ASSET: CurrencyId = DOT;
 const INITIAL_AMOUNT: u128 = 1_000_000_000_000_000;
 const ASSET_ID: u32 = 10;
+const MINIMUM_LIQUIDITY: u128 = 1;
 
 fn assert_last_event<T: Config<I>, I: 'static>(generic_event: <T as Config<I>>::Event) {
     frame_system::Pallet::<T>::assert_last_event(generic_event.into());
@@ -99,7 +100,7 @@ benchmarks_instance_pallet! {
     }: _(
         SystemOrigin::Signed(caller.clone()),
         (BASE_ASSET, QUOTE_ASSET),
-        300_000u128
+        300_000u128 - MINIMUM_LIQUIDITY
     )
     verify {
         assert_last_event::<T, I>(Event::LiquidityRemoved(caller, BASE_ASSET, QUOTE_ASSET).into());
--- a/pallets/amm/src/benchmarking.rs
+++ b/pallets/amm/src/benchmarking.rs
@@ -21,7 +21,7 @@ const BASE_ASSET: CurrencyId = XDOT;
 const QUOTE_ASSET: CurrencyId = DOT;
 const INITIAL_AMOUNT: u128 = 1_000_000_000_000_000;
 const ASSET_ID: u32 = 10;
-const MINIMUM_LIQUIDITY: u128 = 1;
+const MINIMUM_LIQUIDITY: u128 = 1_000u128;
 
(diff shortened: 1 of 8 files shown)


drbh:
@GopherJ what is the best way to replicate sending tokens to `address(0)`? Mining and burning will reduce the total issuance, is there an address we can mint tokens into that is not controlled by anyone?

Current implementation
https://github.com/parallel-finance/parallel/blob/c7772f80103bf24248124cc08a3e66f4a15d80ce/pallets/amm/src/lib.rs#L487-L495

Also please let me know if there are any updates to the logic that are needed 🙏

pallets/amm/src/lib.rs
```diff
@@ -473,8 +480,28 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
         quote_asset: AssetIdOf<T, I>,
         base_amount: BalanceOf<T, I>,
         quote_amount: BalanceOf<T, I>,
+        is_inital_deposit: bool,
     ) -> DispatchResult {
-        T::Assets::mint_into(currency_asset, &who, ownership)?;
+        if is_inital_deposit {
```
GopherJ, line 485: should check `total_isurance` instead

pallets/amm/src/benchmarking.rs
```diff
@@ -21,6 +21,7 @@ const BASE_ASSET: CurrencyId = XDOT;
 const QUOTE_ASSET: CurrencyId = DOT;
 const INITIAL_AMOUNT: u128 = 1_000_000_000_000_000;
 const ASSET_ID: u32 = 10;
+const MINIMUM_LIQUIDITY: u128 = 1;
```
GopherJ, line 24: we can use uniswap's default value

pallets/amm/src/lib.rs
```diff
@@ -91,6 +91,11 @@ pub mod pallet {
         #[pallet::constant]
         type ProtocolFee: Get<Perbill>;
 
+        /// Minimum amount of liquidty needed to init a new pool
+        /// this amount is burned when the pool is created
```
GopherJ, line 95: can you explain also why do we need `MinimumLiquidity`, why this can prevent from attacks in uniswap

pallets/amm/src/lib.rs
```diff
@@ -474,7 +486,30 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
         base_amount: BalanceOf<T, I>,
         quote_amount: BalanceOf<T, I>,
     ) -> DispatchResult {
-        T::Assets::mint_into(currency_asset, &who, ownership)?;
+        // check if any tokens have been issued
+        if T::Assets::total_issuance(currency_asset) == 0 {
```
GopherJ, line 490: is_zero is better

pallets/amm/src/lib.rs
```diff
@@ -474,7 +486,30 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+            let ownership_minus_inital_miniumum_deposit = ownership
+                .checked_sub(T::MinimumLiquidity::get())
+                .ok_or(ArithmeticError::Underflow)?;
+
+            // lock minimum liquidity forever when liquidity is first added
+            T::Assets::mint_into(
+                currency_asset,
```
GopherJ, line 497: can this be a special no private key account like `AccountId([1; 32])`

pallets/router/src/mock.rs
```diff
@@ -132,6 +133,7 @@ impl pallet_amm::Config for Runtime {
     type Event = Event;
     type Assets = CurrencyAdapter;
     type PalletId = AMMPalletId;
+    type LockAccountId = LockAccountId;
```
GopherJ, line 136: I think you can use `RootOperator`, this is already defined

GopherJ:
Otherwise LGTM

GopherJ:
@drbh can you rename `RootOperator` to `OneAccount` which may be better

GopherJ (approved):

PR #1189, opened 2022-01-18
AMM refactoring

ashi20192020ayush:


pallets/amm/src/lib.rs
```diff
@@ -337,11 +336,13 @@ pub mod pallet {
                         .base_amount
                         .checked_sub(base_amount)
                         .ok_or(ArithmeticError::Underflow)?;
+
                     liquidity_amount.quote_amount = liquidity_amount
                         .quote_amount
                         .checked_sub(quote_amount)
                         .ok_or(ArithmeticError::Underflow)?;
 
+                    *pool_liquidity_amount = Some(liquidity_amount);
```
GopherJ, line 345: this is not necessary

pallets/amm/src/lib.rs
```diff
@@ -354,6 +355,7 @@ pub mod pallet {
                                     .quote_amount
                                     .checked_sub(quote_amount)
                                     .ok_or(ArithmeticError::Underflow)?;
+                                *pool_liquidity_amount = Some(*liquidity_amount);
```
GopherJ, line 358: this is not necessary since you are inside `try_mutate`

pallets/amm/src/lib.rs
ashi20192020ayush: @GopherJ I believe, this is necessary otherwise this https://github.com/parallel-finance/parallel/blob/amm-refactoring/pallets/amm/src/tests.rs#L269-L271 would fail.

pallets/amm/src/lib.rs
ashi20192020ayush: replied above

pallets/amm/src/lib.rs
GopherJ: @ashi20192020ayush adding `.as_mut` to https://github.com/parallel-finance/parallel/pull/1189/files#diff-081eb769bf1774964cec3ca16e72ce0510198b2c8e604de2b6b9d1bee336e8ccR317 is better. we can remove line 345

pallets/amm/src/lib.rs
GopherJ: same as above

PR #1194, opened 2022-01-19
Changes to fix #1175

### LINKED ISSUE #1175: No need to record amm liquidity provider's position
liquidity providers already have lp tokens, we dont need to record their position. If the frontend needs it, we can use subql

ashi20192020ayush:
resolves #1175


GopherJ:
LGTM

PR #1197, opened 2022-01-19
fixes #1190

### LINKED ISSUE #1190: avoid AMM calculation overlow when tokens have big decimals
We would like to avoid overflow due to big decimals

ashi20192020ayush:
resolves #1190


ETeissonniere (changes requested):
let's fix the conflicts

PR #1251, opened 2022-02-04
optimize saturating_mul

ashi20192020ayush:


drbh (approved):
lgtm. will merge soon, unless you think we need to make any changes @GopherJ

PR #1252, opened 2022-02-05
Feat/add CODEOWNERS

drbh:
addresses `HAL-04` 

only allow specific developers to update workflows.

update allowed users here: https://github.com/orgs/parallel-finance/teams/workflows-owner


drbh:
@GopherJ please let me know who I should add to the various teams. Thanks

GopherJ (approved):

PR #1260, opened 2022-02-08
simplified local deployment steps

drbh:
adds
- `Makefile` commands for local deployment
- `local-dev.ts` and `yarn local` command to launch directory
- `LOCAL.md` docs


.gitignore
```diff
@@ -32,6 +32,10 @@ dist/
 *.log
 *-local*.json
 
+# resources created for local heiko development
+heiko-dev-para-2085-genesis
```
GopherJ, line 36: this should be more general maybe (not only for heiko & 2085)

Makefile
```diff
@@ -145,5 +145,65 @@ snapshot:
...
+		-d ./resources/cumulus_relay0 \
+		--validator \
+		--alice \
+		--port 50555 \
+		--node-key 0000000000000000000000000000000000000000000000000000000000000001
+
+.PHONY: local-relay-bob
+local-relay-bob:
+	../polkadot/target/release/polkadot \
+		--chain ../polkadot/rococo_local.json \
```
GopherJ, line 168: can we simplify this a bit since we should launch 4 relaychain nodes using `parallelfinance/polkadot`, maybe we can create a for loop to launch them all at once

GopherJ:
The purpose of this PR is for quick debugging, is it possible to add cache for docker build? so that we can reuse the parachain-launch purpose.

If we can make `make image` fast by using cache, then maybe we dont need this pr and it can simplify a lot

PR #1261, opened 2022-02-09
Implemented FixedU128 for AMM reserves

### LINKED ISSUE #1190: avoid AMM calculation overlow when tokens have big decimals

ashi20192020ayush:
resolves #1190


GopherJ:
close in favor of BigUint

PR #1266, opened 2022-02-10
Fix broken yarn lockfile

GopherJ:


GopherJ:
#1267 needs to be addressed seperately

drbh (approved):
🙏 this fixes the issue and now works on the latest version of node `v16.14.0`. Thanks @GopherJ

PR #1278, opened 2022-02-10
adds dockerfile.dev

drbh:
work in progress.

TODO
- [ ] test how if this works when making updates to a pallet
- [ ] use smaller containers for steps 1 and 2

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,46 +1,24 @@
-# This first stage prepares our dependencies to be built by `cargo-chef`.
-FROM docker.io/paritytech/ci-linux:production as planner
-LABEL description="This is the 1st stage: Here we plan the build dependencies with `cargo-chef`."
-
-WORKDIR /parallel
-RUN cargo install cargo-chef
-COPY . .
-RUN cargo chef prepare --recipe-path recipe.json
-
-# This second stage is where the dependencies actually get built.
-# The reason we split it from the first stage is so that the `COPY . .`
-# step doesn't blow our cache.
-FROM docker.io/paritytech/ci-linux:production as cacher
-LABEL description="This is the 2nd stage: In this stage we build the dependencies."
-
-WORKDIR /parallel
-RUN cargo install cargo-chef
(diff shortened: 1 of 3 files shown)


GopherJ:
@drbh we tried cargo-cref before, it doesn't work well. We can try to mount target directory

GopherJ:
You can check rust-musl-builder's example for this. 
image

drbh:
@GopherJ I explored the `-v` option and it's only for `docker run`; [docs](https://docs.docker.com/engine/reference/commandline/run/#options). I think we need a way to cache dependencies when building and it looks like `RUN --mount=type=cache` is a way to do this; [docs](https://docs.docker.com/engine/reference/commandline/run/#options).

I'm having trouble getting this to work on the `parallel` repo - but I can confirm that it works on a dummy repo. Also this solution is mentioned [here](https://github.com/rust-lang/cargo/issues/2644#issuecomment-903929592) and [here](https://stackoverflow.com/a/60590697)

I've pushed the changes to the Dockerfile that I believe should correctly cache the target directory, but I'm getting the following error

computecachekeyissue

I'll continue to explore this solution - but I wanted to share my current findings. Do you have any insight into how we can make this work?

GopherJ:
@drbh maybe we just need to have two dockerfiles:

- one for building parallel binary, mount using `-v` and then use docker run to generate a good binary for second dockerfile
- second to copy parallel binary into image

GopherJ:
@c02fee we can also try polkadot-launch approach, we used it for a while and then replaced it by parachain launch

c02fee:
Bottleneck is with cargo build command 
The binary file cannot be separated to different layers

Other Approaches :- 
* https://github.com/neam/docker-diff-based-layers
* Speed up the cargo build https://github.com/mozilla/sccache

bingryan:
To achieve this purpose, why not consider the wasm way?

GopherJ:
> To achieve this purpose, why not consider the wasm way?

can you elaborate more?

GopherJ:
close for finalizing polkadot work

PR #1280, opened 2022-02-11
simple slowunstake

GopherJ:

close: #1236

### DIFF
--- a/pallets/liquid-staking/src/lib.rs
+++ b/pallets/liquid-staking/src/lib.rs
@@ -46,6 +46,7 @@ pub mod pallet {
         ensure,
         pallet_prelude::*,
         require_transactional,
+        storage::with_transaction,
         traits::{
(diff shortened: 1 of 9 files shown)


pallets/liquid-staking/src/lib.rs
```diff
@@ -241,70 +256,15 @@ pub mod pallet {
...
-
-                // remove unstake request from queue
-                Self::pop_unstake_task()
-            }
-
-            remaining_weight
+            with_transaction(|| match Self::do_pop_front() {
+                Ok(_) => TransactionOutcome::Commit(remaining_weight - base_weight),
+                Err(_) => TransactionOutcome::Rollback(0),
+            })
```
MrPai, line 267: Seems like now we only process one account each time, due to the `unstake queue cap` and `minunstake amount`, maybe this would be the throughput for the system.

MrPai (approved):
There would be one concern and others look good for me.

PR #1309, opened 2022-02-14
Feat/update event syntax

drbh:
use literal event structure and added a helper function `calculate_reserves_to_remove`

### DIFF
--- a/pallets/amm/src/benchmarking.rs
+++ b/pallets/amm/src/benchmarking.rs
@@ -86,16 +86,16 @@ benchmarks_instance_pallet! {
         (5u128, 5u128)
     )
     verify {
-        assert_last_event::<T, I>(Event::<T, I>::LiquidityAdded {
-            sender: caller,
-            base_currency_id: BASE_ASSET,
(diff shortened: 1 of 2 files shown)


GopherJ:
hi @drbh what's the context of this one? Because the syntax is global, feel like if we would like to change, we should change for all pallets to make things consistent.

In TypeScript we should be able to do `const [..] = event` for having all those fields

drbh:
Hey @GopherJ I only made the change because it makes the events easier to read in the code. I think we may want to use this syntax in all of the pallets but that can be a separate bigger refactor PR. Is there any reason we should not use the literal syntax?

I've removed the new syntax from this PR so we can merge the changes to the events. Thanks!

GopherJ:
@drbh I just checked substrate it seems they also didn't use the new syntax yet. I think the current syntax is also ok as long as we have comments for every parameter.

drbh:
@GopherJ okay sounds good. Thanks for checking 🙏

Please let me know if I should change anything else about the PR before merging

GopherJ (approved):

Issue #1283, opened 2022-02-11
Extract the coversion between staking_currency_amount & liquid_currency_amount into functions


GopherJ:
https://github.com/parallel-finance/parallel/blob/master/pallets/liquid-staking/src/types.rs#L33-L35
https://github.com/parallel-finance/parallel/blob/master/pallets/liquid-staking/src/lib.rs#L306-L309
https://github.com/parallel-finance/parallel/blob/master/pallets/liquid-staking/src/lib.rs#L355-L357

PR #1311, opened 2022-02-15
refactor staking/liquid amount conversion

GopherJ:

close: #1283

### DIFF
--- a/pallets/liquid-staking/src/lib.rs
+++ b/pallets/liquid-staking/src/lib.rs
@@ -33,10 +33,12 @@ pub mod weights;
 #[macro_use]
 extern crate primitives;
 
-use frame_support::traits::{fungibles::InspectMetadata, Get};
+use frame_support::traits::{fungibles::InspectMetadata, tokens::Balance as BalanceT, Get};
 use pallet_xcm::ensure_response;
-use primitives::{ExchangeRateProvider, LiquidStakingCurrenciesProvider, Rate};
-use sp_runtime::traits::Zero;
+use primitives::{
+    ExchangeRateProvider, LiquidStakingConvert, LiquidStakingCurrenciesProvider, Rate,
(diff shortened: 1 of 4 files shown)


pallets/liquid-staking/src/lib.rs
```diff
@@ -788,6 +771,27 @@ pub mod pallet {
...
+        fn do_update_exchange_rate(bonding_amount: BalanceOf<T>) -> DispatchResult {
+            let old_exchange_rate = Self::exchange_rate();
+            let matching_pool = Self::matching_pool();
+            match Rate::checked_from_rational(
+                bonding_amount
+                    .checked_add(matching_pool.total_stake_amount)
+                    .ok_or(ArithmeticError::Overflow)?,
+                T::Assets::total_issuance(Self::liquid_currency()?)
+                    .checked_add(matching_pool.total_unstake_amount)
+                    .ok_or(ArithmeticError::Overflow)?,
```
GopherJ, line 784: @MrPai  will it cause issues if we use the edited matching_ledger to update exchange_rate? if yes I'll still need to pass old matching_pool to this function

MrPai: We didn't think about what was matched

MrPai (commented):
others looks good to me.

pallets/liquid-staking/src/lib.rs
GopherJ: sorry can you elaborate more?

GopherJ: ok so you still want to use old matching_ledger right?

MrPai: `old_matching_ledger` looks clear for now, however  edited matching_ledger maybe confused.

PR #1323, opened 2022-02-16
Added set_lock and remove_lock

### LINKED ISSUE #1276: add force_set_lock, force_remove_lock call to currency-adapter
At the beginning, we just check if the `AssetId` is 0 OR 1, if not then return Error. If yes, we can then use pallet_balances's set_lock & remove_lock

ashi20192020ayush:
resolves #1276


GopherJ (commented):
it should be dispatchable calls @ashi20192020ayush

PR #1339, opened 2022-02-18
Fixed issue -> calling the function with non existing pairs will cause the code to panic

c02fee:
Addresses the issue 

Thanks for the update , during writing some custom test cases it was observed calling the function with non existing pairs will cause the code to panic.


drbh (approved):
lgtm

PR #1352, opened 2022-02-22
Handled Panic event in get_amounts_in

c02fee:
Fixed Panic when `// Test case for Panic when amount_out >= supply_out`
Otherwise will Panic with error `ArithmeticError::Overflow`


drbh (approved):
lgtm

PR #1364, opened 2022-02-23
impl stable swap math

drbh:
WORK IN PROGRESS

implementing stable swap math with help from this article https://miguelmota.com/blog/understanding-stableswap-curve/ and this  https://github.com/curvefi/curve-contract/blob/master/contracts/pool-templates/base/SwapTemplateBase.vy contract.

TODO 
- [X] Write initial `get_d` function
- [X] Write initial `get_y` function
- [ ] Convert all floats to FixedU128
- [ ] Review math (compare with other impls)
- [ ] Maybe add function to ramp up/down `A`
- [ ] Review `(add|remove)_liquidity` functions for stable swap logic
- [ ] Design a way to swap between curves (from config)
- [ ] Move constants into config

Open questions
- How do we plan on handling ramping up/down A?
- How do we plan on supporting new curves? On AMM instantiation? 
- We may have to handle changes the `liquidity` and `swap` how does this change the design?

### DIFF
--- a/pallets/amm/src/lib.rs
+++ b/pallets/amm/src/lib.rs
@@ -618,23 +618,26 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
     #[allow(dead_code)]
     fn get_d(
         (asset_in, asset_out): (AssetIdOf<T, I>, AssetIdOf<T, I>),
-    ) -> Result<f64, DispatchError> {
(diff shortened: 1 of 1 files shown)


drbh:
closing in favor of https://github.com/parallel-finance/parallel/pull/1515

PR #1389, opened 2022-02-28
correct liquidity logic

drbh:
liquidity (LP token) amount was incorrectly calculated because we updated reserves before calculating ratio.

This PR moves the logic below the calculation so the correct amount is returned.


drbh:
### Reference

This should now match the Uniswap V2 logic here: https://github.com/Uniswap/v2-core/blob/4dd59067c76dea4a0e8e4bfdda41877a6b16dedc/contracts/UniswapV2Pair.sol#L110-L131
uniswapliqlogic

ETeissonniere (approved):

PR #1399, opened 2022-03-02
lease period

alannotnerd:
- Change leasePeriod type alias
- Change leasePeriod type alias


primitives/src/lib.rs
```diff
@@ -89,7 +89,7 @@ pub type BridgeId = (ChainNonce, ChainNonce);
 
 pub const SECONDS_PER_YEAR: Timestamp = 365 * 24 * 60 * 60;
 
-pub type LeasePeriod = BlockNumber;
+pub type LeasePeriod = u32;
```
GopherJ, line 92: on relaychain `LeasePeriod` is BlockNumber, and we use the same BlockNumber like in relaychain (u32), so it should be fine maybe?

PR #1408, opened 2022-03-03
Upgrade to Polkadot.0.9.17

ashi20192020ayush:


GopherJ:
let's freeze this util 17/03

pallets/liquid-staking/src/mock.rs
```diff
@@ -129,12 +133,37 @@ impl Config for XcmConfig {
     type AssetClaims = PolkadotXcm;
 }
 
+pub struct SystemParachainAsSuperuser<Origin>(PhantomData<Origin>);
```
rocketman1997, line 136: duplicate several times, maybe we can refact a bit clean

rocketman1997: also we may not really need such superuser control from other parachain, we can do it from relaychain if nessessary

ashi20192020ayush: This is mock config which we need for cumulus_pallet_xcmp_queue

PR #1523, opened 2022-03-22
convert u128 to allowed FixedU128 return type

drbh:
RPC panics on u128 serialization. However FixedU128s are allowed

### DIFF
--- a/pallets/router/rpc/runtime-api/src/lib.rs
+++ b/pallets/router/rpc/runtime-api/src/lib.rs
@@ -16,8 +16,7 @@
 
 use codec::Codec;
 use primitives::{Balance, CurrencyId};
-use sp_runtime::DispatchError;
-use sp_runtime::FixedU128;
+use sp_runtime::{DispatchError, FixedU128};
 use sp_std::vec::Vec;
 
 sp_api::decl_runtime_apis! {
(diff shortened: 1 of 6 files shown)


drbh:
@GopherJ this PR fixes the issue with the `getBestRoute` RPC call. Currently the `u128` return type will cause the node to panic. Using a `FixedU128` resolves this issue and we should merge and deploy this ASAP to avoid any downtime issues

drbh:
for reference this is the error that was occurring 
```
Thread 'tokio-runtime-worker' panicked at 'Expected always-serializable type; qed: Error("u128 is not supported", line: 0, column: 0)'
```

drbh:
related frontend changes https://github.com/parallel-finance/parallel-js/pull/34

GopherJ (approved):

GopherJ:
@drbh looks in parallel-js it should return `FixedU128` instead of `Balance` right?

rocketman1997:
@drbh curios just change to `FixedU128` will resolve the serializable issue? Or should we wrap u128 as `NumberOrHex` like [this](https://github.com/paritytech/substrate/blob/22d40c761a985482f93bbbea5ba4199bdba74f8e/frame/transaction-payment/rpc/src/lib.rs#L144)

PR #1555, opened 2022-04-01
Reedem amount check

sumaninster9:
Reedem zero amount issue is fixed


pallets/loans/src/lib.rs
```diff
@@ -180,6 +180,8 @@ pub mod pallet {
         InvalidAmount,
         /// Payer cannot be signer
         PayerIsSigner,
+		//Redeem Amount cannot be zero
+		InsufficientRedeemAmount,
```
GopherJ, line 184: make fmt is needed

pallets/loans/src/lib.rs
```diff
@@ -650,6 +652,7 @@ pub mod pallet {
             asset_id: AssetIdOf<T>,
             #[pallet::compact] redeem_amount: BalanceOf<T>,
         ) -> DispatchResultWithPostInfo {
+			ensure!(redeem_amount > 0, Error::<T>::InsufficientRedeemAmount);
```
GopherJ, line 655: should reuse `InvalidAmount` and probably !redeem_amount.is_zero() is better

sumaninster9:
> Hey @sumaninster9, thanks for the PR! just a few comments
> 
> 1. The `fix/add-zero-redeem-check` includes the changes needed for both the zero redeem check, and the check that the pallet contains enough liquidity. Generally a PR should only include a single feature. I apologize for not clarifying this earlier, but going forward we should open separate PRs for separate fix and features.
> 2. Please make sure to not include any extra files/changes in the PR. It looks like some comments were added into the `yarn.lock` file that are not related directly to this PR.
> 3. After resolving comments please makes sure to click `Resolve Conversation` and re request an update from the person who's left the comments.
> 
> Overall the logic and tests look good - these comments are just related to Parallel's development workflow

Thanks. Yes, by mistake I put both the fixes in single PR. I will be careful of that next time. When I run 'make init', that time possibly something got written to 'yarn.lock'. Okay I will do "Resolve Conversation" . Thank you again for all the clarifications and suggestions.

GopherJ (approved):

PR #1560, opened 2022-04-04
accrue interest with user action

yz89:


pallets/loans/src/lib.rs
```diff
@@ -232,8 +231,9 @@ pub mod pallet {
 
     /// The timestamp of the last calculation of accrued interest
     #[pallet::storage]
-    #[pallet::getter(fn last_accrued_timestamp)]
-    pub type LastAccruedTimestamp<T: Config> = StorageValue<_, Timestamp, ValueQuery>;
+    #[pallet::getter(fn last_accrued_interest_time)]
+    pub type LastAccruedInterestTime<T: Config> =
```
GopherJ, line 235: will we have migration? but this looks much more clear

pallets/loans/src/interest.rs
```diff
@@ -20,28 +20,36 @@ use sp_runtime::{traits::Zero, DispatchResult};
...
-                InterestRateModel::get_supply_rate(borrow_rate, util, market.reserve_factor);
-
-            UtilizationRatio::<T>::insert(asset_id, util);
-            BorrowRate::<T>::insert(asset_id, &borrow_rate);
-            SupplyRate::<T>::insert(asset_id, supply_rate);
-
-            Self::update_borrow_index(borrow_rate, asset_id, &market, delta_time)?;
-            Self::update_exchange_rate(asset_id)?;
+    /// Accrue interest and update corresponding storage
+    pub(crate) fn accrue_interest(asset_id: AssetIdOf<T>) -> DispatchResult {
```
GopherJ, line 24: probably `#[require_transactional]` can be added to avoid usage in non-transactional function

GopherJ:
looks great, want this feature for long time:)

GopherJ (approved):
probably need migration on lastAccuredTime to avoid incorrect delta time, otherwise LGTM

pallets/loans/src/interest.rs
```diff
@@ -20,28 +20,36 @@ use sp_runtime::{traits::Zero, DispatchResult};
...
+        let now = T::UnixTime::now().as_secs();
+        let last_accrued_interest_time = Self::last_accrued_interest_time(asset_id);
+        if last_accrued_interest_time.is_zero() {
+            // For the initialization
+            Self::update_last_accrued_interest_time(asset_id, now)?;
```
GopherJ, line 29: seems we are duplicating this line twice, should probably do it once in line 35 and write only:

```
if last_accrued_interest_time.is_zero() {
  return Ok(());
}
```

pallets/loans/src/lib.rs
yz89: we don't need to migrate this for we just use new storage here.

zhoujia6139:
LGTM and yeah I think we don't need migration for lastAccuredTime because we just remove an old storage and add a new one.  We can refer to [this link of substrate doc](https://docs.substrate.io/v3/advanced/storage/)

PR #1574, opened 2022-04-08
Refactored delta calculation

c02fee:
Existing delta calculation happens from reserved values. 
Calculation requires  newly added values which is not persisted yet in add_liquidity_pool function

Kept existing functionality and introduced a new functionality which calculates using the persisted values + new values


drbh (approved):
LGTM ✅

PR #1605, opened 2022-04-18
ls/migration

MrPai:


pallets/liquid-staking/src/migrations.rs
```diff
@@ -13,20 +13,81 @@
...
     }
 
     pub fn migrate<T: Config>() -> Weight {
-        0
+        if StorageVersion::<T>::get() == crate::Versions::V2 {
+            log::info!("migrating liquidStaking to Versions::V3",);
+            // 1. from MarketCap to StakingLedgerCap
+            let cap = MarketCap::get();
+            MarketCap::kill();
+            StakingLedgerCap::<T>::put(cap);
```
GopherJ, line 51: since we need to do migration storage version v3 for both heiko & parallel and they have different staking ledger cap, I think we can left it to be empty and set it once runtime upgrade succeeded

pallets/liquid-staking/src/migrations.rs
```diff
@@ -13,20 +13,81 @@
...
+
+            // 2. update MatchingPool, MatchingLedger
+            let r = MatchingPool::<T>::translate::<OldMatchingLedger<BalanceOf<T>>, _>(
```
GopherJ, line 54: this needs to be working in parallel

MrPai: It seems like that data on Heiko needs to be migrated, parallel is already the newest data structure.

MrPai:
@GopherJ  Tests works now, let's keep in mind:
- Do runtime upgrade once relaychain updates its era, 
- and manually set the `stakingLegdeCap` a proper value once upgrade success.

pallets/liquid-staking/src/migrations.rs
GopherJ: yes but the storageVersion on parallel needs to be set

GopherJ (approved):

GopherJ:
> @GopherJ Tests works now, let's keep in mind:
> 
>     * Do runtime upgrade once relaychain updates its era,
> 
>     * and manually set the `stakingLegdeCap` a proper value once upgrade success.

sure will apply this logic

PR #1606, opened 2022-04-18
Feat/maths rs

c02fee:


c02fee:
Addressed via :- https://github.com/parallel-finance/parallel/pull/1633

PR #1610, opened 2022-04-18
fix protocol fee and fix for linter

drbh:
update the protocol fee to use `delta_util` instead of the square root of k. Also fixed some linting issues


c02fee:
Good to be merged !

PR #1614, opened 2022-04-19
sibling parachain token transfer integration tests

ashi20192020ayush:


integration-tests/src/kusama_test_net.rs
```diff
@@ -41,6 +41,16 @@ decl_test_parachain! {
     }
 }
 
+decl_test_parachain! {
+    pub struct Karura {
+        Runtime = karura_runtime::Runtime,
+        Origin = karura_runtime::Origin,
+        XcmpMessageHandler = karura_runtime::XcmpQueue,
+        DmpMessageHandler = karura_runtime::DmpQueue,
```
GopherJ, line 49: need real token transfer tests

PR #1623, opened 2022-04-19
make amm aware of native tokens

drbh:
currently the `keep_alive` check prohibits users from transferring their full balances when swapping. This causes an error with the "Max" button on the frontend and results in dust.

This PR adds the `GetNativeCurrencyId` to the AMM so we can check if a user is attempting to transfer native tokens. We should only use the `keep_alive` flag on native balances and set it false for all assets. This way users can transfer 100% of their assets that are not native

### DIFF
--- a/pallets/amm/src/lib.rs
+++ b/pallets/amm/src/lib.rs
@@ -120,7 +120,7 @@ pub mod pallet {
         type MaxLengthRoute: Get<u32>;
 
         #[pallet::constant]
-        type GetNativeCurrencyId: Get<u32>;
+        type GetNativeCurrencyId: Get<AssetIdOf<Self, I>>;
     }
 
     #[pallet::error]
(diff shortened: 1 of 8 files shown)


pallets/amm/src/lib.rs
```diff
@@ -865,8 +868,20 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+        T::Assets::transfer(
+            base_asset,
+            &Self::account_id(),
+            who,
+            base_amount,
+            base_asset == T::GetNativeCurrencyId::get(), // should keep alive if is native
+        )?;
+        T::Assets::transfer(
+            quote_asset,
+            &Self::account_id(),
```
GopherJ, line 880: do we need to keep alive pallet account as well? in that case, there will be 0.01HKO in pool which cannot be swapped, but that's not an issue I guess

GopherJ:
otherwise looks good!

GopherJ:
similar thing should also be done in frontend for calculating the Max value I think

GopherJ (approved):

pallets/amm/src/lib.rs
drbh: yea I wasn't sure about that. my initial concern was that we could accidentally delete the pallet account from the chain while it still has assets. So I left the keep_alive check on since I don't think it will cause issues

PR #1636, opened 2022-04-23
add rpc interface to get loans market status.

zhoujia6139:


pallets/loans/rpc/runtime-api/src/lib.rs
```diff
@@ -15,12 +15,13 @@
...
-use primitives::{Liquidity, Shortfall};
-use sp_runtime::DispatchError;
+use primitives::{CurrencyId, Liquidity, Rate, Ratio, Shortfall};
+use sp_runtime::{DispatchError, FixedU128};
 
 sp_api::decl_runtime_apis! {
     pub trait LoansApi<AccountId> where
         AccountId: Codec, {
         fn get_account_liquidity(account: AccountId) -> Result<(Liquidity, Shortfall), DispatchError>;
+        fn get_market_status(asset_id: CurrencyId) -> Result<(Rate, Rate, Rate, Ratio, u128, u128, FixedU128), DispatchError>;
```
yrong, line 25: u128 may not be safe as rpc return

GopherJ: we have same issue in amm router I guess, we can check getBestRoute rpc

zhoujia6139: 👌🏻

pallets/loans/src/lib.rs
```diff
@@ -785,11 +785,12 @@ pub mod pallet {
             Self::update_reward_supply_index(asset_id)?;
             Self::distribute_supplier_reward(asset_id, &who)?;
 
-            T::Assets::transfer(asset_id, &who, &Self::account_id(), mint_amount, false)?;
-            Self::update_earned_stored(&who, asset_id)?;
-            let exchange_rate = Self::exchange_rate(asset_id);
+            let exchange_rate = Self::exchange_rate_stored(asset_id)?;
+            Self::update_earned_stored(&who, asset_id, exchange_rate)?;
```
GopherJ, line 789: shall it be `earned_stored`

pallets/loans/src/interest.rs
```diff
@@ -32,52 +32,105 @@ impl<T: Config> Pallet<T> {
...
-        UtilizationRatio::<T>::insert(asset_id, util);
-        BorrowRate::<T>::insert(asset_id, &borrow_rate);
-        SupplyRate::<T>::insert(asset_id, supply_rate);
-        Self::update_borrow_index(borrow_rate, asset_id, &market, delta_time)?;
-        Self::update_exchange_rate(asset_id)?;
 
-        Ok(())
+        let now = T::UnixTime::now().as_secs();
+        let last_accrued_interest_time = Self::last_accrued_interest_time(asset_id);
+        if now > last_accrued_interest_time {
```
yz89, line 91: I think we should return Err once `now <= last_accrued_interest_time` rather than continue calculate the exchange_rate.

zhoujia6139: get_market_status is just read_only function. And if we return Err, that means if accruedInterest happened in Block 1000, and we can not fetch the status at Block 1000. That's not reasonable.

pallets/loans/src/lib.rs
yz89: Do you mean to rename update_earned_stored to earned_stored?

pallets/loans/src/interest.rs
yz89: OK, got it.

yz89 (approved):

PR #1692, opened 2022-05-02
revert reward accured migration.

zhoujia6139:


pallets/loans/src/lib.rs
```diff
@@ -448,6 +454,28 @@ pub mod pallet {
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
+        ///Migrate reward data from RewardAccured to RewardAccrued
+        #[pallet::weight(T::WeightInfo::migrate_reward_data())]
+        #[transactional]
+        pub fn migrate_reward_data(_origin: OriginFor<T>) -> DispatchResultWithPostInfo {
```
yrong, line 460: better to make sure this call can be executed by only someone

pallets/loans/src/lib.rs
```diff
@@ -448,6 +454,28 @@ pub mod pallet {
...
+            let mut remove_keys = Vec::new();
+            for (account_id, reward_amount) in RewardAccured::<T>::iter() {
+                if reward_amount > 0 {
+                    RewardAccrued::<T>::insert(account_id.clone(), reward_amount);
+                }
+                remove_keys.push(account_id);
+                if remove_keys.len() >= 30 {
```
yrong, line 467: I would assume 100 should be fine

pallets/loans/src/lib.rs
```diff
@@ -448,6 +454,28 @@ pub mod pallet {
...
+                    break;
+                }
+            }
+
+            for key in remove_keys.iter() {
+                RewardAccured::<T>::remove(key);
+            }
+
+            Ok(().into())
```
yrong, line 476: better to emit some event to tell when the migration finished

pallets/loans/src/lib.rs
GopherJ, line 467: we have many zero, guess we shouldn't use `remove_keys` length as the break point, because 0 has been iterated => consumed block weights

zhoujia6139: Zero reward account has been pushed into remove_keys vector. So 30 is the max iteration count.

pallets/loans/src/lib.rs
zhoujia6139: Bot can check remaining amount before send a `migrate_reward_data` transaction.

yz89 (approved):

PR #1699, opened 2022-05-03
Feat/streaming storage

rjman-ljm:
## Context

On-chain storage is a simple way to query the historical account streams
Off-chain endpoint/service cannot not update immediately due to delay of subquery and db read/write.

- Senders: <AccountId, Vec\<StreamId>>
- Receivers: <AccountId, Vec\<StreamId>>


pallets/streaming/src/lib.rs
```diff
@@ -55,7 +55,12 @@ type BalanceOf<T> =
     <<T as Config>::Assets as Inspect<<T as frame_system::Config>::AccountId>>::Balance;
 type AccountOf<T> = <T as frame_system::Config>::AccountId;
 
-// Struct for Stream
+#[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub enum StreamStatus {
+    Ongoing,
+    Completed,
```
GopherJ, line 61: why do we need this field? if there aren't any funds left then certainly it's Completed

pallets/streaming/src/lib.rs
```diff
@@ -76,13 +81,16 @@ pub struct Stream<T: Config> {
...
+    // The current status of the stream
+    status: StreamStatus,
 }
 
 #[frame_support::pallet]
 pub mod pallet {
     use super::*;
 
     use weights::WeightInfo;
+    pub const MAX_STREAM_LEN: usize = 1024;
```
GopherJ, line 93: probably too big, I would suggest 32 to start

pallets/streaming/src/lib.rs
```diff
@@ -161,7 +173,17 @@ pub mod pallet {
     /// Global storage for streams
     #[pallet::storage]
     #[pallet::getter(fn get_stream)]
-    pub type Streams<T: Config> = StorageMap<_, Twox64Concat, StreamId, Stream<T>>;
+    pub type Streams<T: Config> = StorageMap<_, Blake2_128Concat, StreamId, Stream<T>, OptionQuery>;
+
+    #[pallet::storage]
+    #[pallet::getter(fn senders)]
+    pub type Senders<T: Config> =
+        StorageMap<_, Blake2_128Concat, T::AccountId, Vec<StreamId>, OptionQuery>;
```
GopherJ, line 181: BoundedVec preferred

pallets/streaming/src/lib.rs
```diff
@@ -161,7 +173,17 @@ pub mod pallet {
...
+
+    #[pallet::storage]
+    #[pallet::getter(fn recipients)]
+    pub type Recipients<T: Config> =
+        StorageMap<_, Blake2_128Concat, T::AccountId, Vec<StreamId>, OptionQuery>;
```
GopherJ, line 186: BoundedVec preferred

pallets/streaming/src/lib.rs
```diff
@@ -161,7 +173,17 @@ pub mod pallet {
     /// Global storage for streams
     #[pallet::storage]
     #[pallet::getter(fn get_stream)]
-    pub type Streams<T: Config> = StorageMap<_, Twox64Concat, StreamId, Stream<T>>;
+    pub type Streams<T: Config> = StorageMap<_, Blake2_128Concat, StreamId, Stream<T>, OptionQuery>;
```
GopherJ, line 176: Feel we just need StorageDoubleMap: sender -> receiver -> Vec<Stream>

GopherJ (changes requested):
Feel data structure can still be improved

pallets/streaming/src/lib.rs
rjman-ljm: If combined with the loans pallet later, I think there will be more status, @i5possible what do you think of this?

pallets/streaming/src/lib.rs
rjman-ljm: okay

pallets/streaming/src/lib.rs
rjman-ljm: This is a front-end requirement, we need to query the scent and received streams based on AccountId respectively.

pallets/streaming/src/lib.rs
rjman-ljm: yep

pallets/streaming/src/lib.rs
i5possible: I am thinking of the Cancelled/AsCollatoral/Redeemed status. If we don't have the status now. We may need to have a data migration. Any other thoughts?

runtime/parallel/src/lib.rs
```diff
@@ -1773,12 +1773,14 @@ impl pallet_crowdloans::Config for Runtime {
 
 parameter_types! {
     pub const StreamPalletId: PalletId = PalletId(*b"par/strm");
+    pub const MaxStreamsCount: u32 = 32;
```
i5possible, line 1776: Not sure if it's true. If the user is an account in a company, he wants to send the streams for their employee. I believe there will be more than 32 streams?

pallets/streaming/src/lib.rs
i5possible: I think CJ means that we could use a StorageDoubleMap so that we could query either by the sender or receiver to save the storage?

rjman-ljm: I understand it, CJ means that the keys of this two storage are different, but they maintain the same values, so it can be optimized by using double keys as the storage structure. I agree with this, but the requirement from the front end makes it impossible to use `StorageDoubleMap `. Take `<AssetId, AccountId> -> Balance` as an example, we must provide both keys in order to query the corresponding value, but when we query a receive stream, we don't care who is the sender.

runtime/parallel/src/lib.rs
rjman-ljm: Maybe we need to discuss it.

pallets/streaming/src/lib.rs
i5possible: Yeah for sure.I think you are right about it.

GopherJ: @rjman-ljm how about `StorageNMap : assetId -> sender -> receiver -> Vec`

pallets/streaming/src/lib.rs
GopherJ: If we cancel then the stream is removed actually so the status wont be useful

pallets/streaming/src/lib.rs
GopherJ: though I'm not sure if the frontend is easy to get received streams

runtime/parallel/src/lib.rs
GopherJ: this value is easy to change, we can adjust in the future, but for now I think 32 is very enough actually

pallets/streaming/src/lib.rs
rjman-ljm: Three keys? I think it would be unfriendly for front-end to query If both sender and receive are keys.

rjman-ljm: How about `StorageDoubleMap: AccountId -> StreamKind(Send | Receive) -> StreamId(u32)`?

runtime/parallel/src/lib.rs
i5possible: 128 Should be good to start. I think we will have more than 32 from our account. Confirmed with @Helen8

rjman-ljm: okay

GopherJ (approved):

PR #1725, opened 2022-05-09
fix creating unpayable stream && rename call name

rjman-ljm:


pallets/streaming/src/lib.rs
```diff
@@ -100,17 +100,19 @@ pub mod pallet {
...
+        /// The rate per second calculated is zero
+        InvalidRatePerSecond,
         /// The stream id is not found
         InvalidStreamId,
-        /// Caller is not the streamer
-        NotTheStreamer,
-        /// Caller is not the recipient
-        NotTheRecipient,
+        /// Caller is not the stream sender
+        NotTheStreamSender,
```
GopherJ, line 113: can be simplified as:

- NotTheSender
- NotTheRecipient

GopherJ (approved):

pallets/streaming/src/lib.rs
rjman-ljm: yep

GopherJ:
can you rename also to:

- create_stream -> create
- withdraw_from_stream -> withdraw
- cancel_stream -> cancel

cc @i5possible

rjman-ljm:
> can you rename also to:
> 
> * create_stream -> create
> * withdraw_from_stream -> withdraw
> * cancel_stream -> cancel
> 
> cc @i5possible
I think it would be better

rjman-ljm:
@GopherJ  I tried it and found it would change a lot, maybe do it in the next PR.

PR #1727, opened 2022-05-09
expand stream structure for loans pallet

rjman-ljm:


pallets/streaming/src/types.rs
```diff
@@ -14,6 +14,8 @@ pub enum StreamStatus {
     Completed,
```
GopherJ, line 14: should put inside Ongoing

rjman-ljm: Why do it, I don't understand

GopherJ: how about Ongoing { asCollateral: false|true }

rjman-ljm:
@GopherJ I've pushed some changes

rjman-ljm:
1. current implementation
2. struct Ongoing { asCollateral: false | true } &&  struct Completed { cancelled: false | true }

Which one is better

rjman-ljm:
@GopherJ I've pushed some changes

pallets/streaming/src/types.rs
rjman-ljm: okay

pallets/streaming/src/types.rs
```diff
@@ -70,9 +75,13 @@ impl<T: Config> Stream<T> {
             recipient,
             start_time,
             end_time,
-            status: StreamStatus::Ongoing,
+            status: StreamStatus::Ongoing {
+                as_collateral: false,
+            },
+            cancellable: true,
```
i5possible, line 81: Should we add it to the create method and set the default true if not provided?

rjman-ljm: If we support creating a non-cancellable stream, we need to add a new field to create(), and we must pass in field

i5possible (approved):
LGTM

i5possible:
error: linking with `cc` failed: exit status: 1

Could you compile it in local?

pallets/streaming/src/types.rs
rjman-ljm: Already added

rjman-ljm:
@GopherJ The PR for fixing CI was merged, please do a review for these changes

GopherJ (approved):

PR #1737, opened 2022-05-10
Changes to remove notify parameter from XCM proxy calls

ashi20192020ayush:


pallets/xcm-helper/src/lib.rs
```diff
@@ -576,34 +571,25 @@ impl<T: Config> XcmHelper<T, BalanceOf<T>, AccountIdOf<T>> for Pallet<T> {
...
         }))
     }
 
     fn do_remove_proxy(
         delegate: AccountId,
         proxy_type: Option<ProxyType>,
         delay: BlockNumber,
-        notify: impl Into<<T as pallet_xcm::Config>::Call>,
-    ) -> Result<QueryId, DispatchError> {
+    ) -> Result<Option<QueryId>, DispatchError> {
```
GopherJ, line 592: same as above

pallets/xcm-helper/src/lib.rs
```diff
@@ -468,15 +466,13 @@ pub trait XcmHelper<T: pallet_xcm::Config, Balance, TAccountId> {
...
-    ) -> Result<QueryId, DispatchError>;
+    ) -> Result<Option<QueryId>, DispatchError>;
 
     fn do_remove_proxy(
         delegate: AccountId,
         proxy_type: Option<ProxyType>,
         delay: BlockNumber,
-        notify: impl Into<<T as pallet_xcm::Config>::Call>,
-    ) -> Result<QueryId, DispatchError>;
+    ) -> Result<Option<QueryId>, DispatchError>;
```
GopherJ, line 475: same

pallets/xcm-helper/src/lib.rs
```diff
@@ -468,15 +466,13 @@ pub trait XcmHelper<T: pallet_xcm::Config, Balance, TAccountId> {
...
```
GopherJ, line 469: same as below

GopherJ:
we also need to make `notify` parameter an `Option` for all `do_*` helper functions so that it can applies to two cases

PR #1742, opened 2022-05-11
Liquidation free loans

alannotnerd:
Background:
The back collateral will be liquidated if the price of target/collateral is bumpy. The PR amis to introduce a liquidation-free way for loans. The LFL mode is based on the relationship of target and collateral asset. Regularly they're strongly relative with each other, and we'll give a volation for collateral which ensure itself covers the principle and intrest.

Revamp including:
1. Add a mapping from the target asset to the allowed back collateral in liquidation free mode, which's collateral factor must be ZERO in normal loans market.
2. Includes the collateral value regardless the collateral factor when borrowing target asset in that map.

TODO:
- [ ] redeem
- [ ] repay borrow
- [ ] liquidate borrow


pallets/loans/src/lib.rs
```diff
@@ -1164,6 +1179,31 @@ impl<T: Config> Pallet<T> {
         T::PalletId::get().into_account()
     }
 
+    fn get_dot_base_position(account: &T::AccountId) -> Result<FixedU128, DispatchError> {
```
yrong, line 1182: better not name as `dot***`,  in kusama it should be ksm based

pallets/loans/src/lib.rs
```diff
@@ -1164,6 +1179,31 @@ impl<T: Config> Pallet<T> {
...
+        {
+            total_asset_value = total_asset_value
+                .checked_add(&Self::collateral_asset_value(account, asset_id)?)
+                .ok_or(ArithmeticError::Overflow)?;
+        }
+        Ok(total_asset_value)
+    }
+
+    fn get_account_lf_liquidity(account: &T::AccountId) -> Result<Liquidity, DispatchError> {
+        let dot_borrowed_amount = Self::current_borrow_balance(account, DOT)?;
```
yrong, line 1195: same as above

pallets/loans/src/lib.rs
```diff
@@ -260,6 +266,11 @@ pub mod pallet {
     pub type LastAccruedInterestTime<T: Config> =
         StorageMap<_, Blake2_128Concat, AssetIdOf<T>, Timestamp, ValueQuery>;
 
+    /// Liquidation free collateral.
+    #[pallet::storage]
+    #[pallet::getter(fn liquidation_free_collaterals)]
+    pub type LiquidationFreeCollaterals<T: Config> = StorageValue<_, Vec<AssetIdOf<T>>, ValueQuery>;
```
yrong, line 272: may not need this storage

alannotnerd: Why?

alannotnerd: 1. The vault cannot be dissolved when ctoken haven't redeemed. We cannot liquidate ctoken when the vault exists.
2. The ctoken cannot be force redeemed because the owner is our pallet when ctoken is collateral.

Due to 1) and 2), we cannot use vault as equivalent of the storage.

pallets/loans/src/lib.rs
```diff
@@ -22,6 +22,8 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+use core::cmp::max;
```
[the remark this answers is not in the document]
alannotnerd: it doesn't under std namespace

pallets/loans/src/lib.rs
```diff
@@ -1784,13 +1855,25 @@ impl<T: Config> Pallet<T> {
...
-        if liquidity < reduce_amount {
-            Err(Error::<T>::InsufficientLiquidity.into())
-        } else {
-            Ok(())
+    // `dot_lf`: check in liquidation free mode which means borrowing dot or redeeming assets in
+    // `LiquidationFreeCollaterals`.
+    fn ensure_liquidity(
+        account: &T::AccountId,
+        reduce_amount: FixedU128,
+        dot_lf: bool,
```
yrong, line 1863: better rename as `lf_enabled`

pallets/loans/src/lib.rs
```diff
@@ -1784,13 +1855,25 @@ impl<T: Config> Pallet<T> {
...
+    ) -> DispatchResult {
+        let (total_liquidity, _) = Self::get_account_liquidity(account)?;
+        let lf_liquidity = Self::get_account_lf_liquidity(account)?;
+
+        if dot_lf && max(total_liquidity, lf_liquidity) > reduce_amount {
```
yrong, line 1868: ditto

pallets/loans/src/lib.rs
```diff
@@ -1784,13 +1855,25 @@ impl<T: Config> Pallet<T> {
...
-            Ok(())
+    // `dot_lf`: check in liquidation free mode which means borrowing dot or redeeming assets in
+    // `LiquidationFreeCollaterals`.
+    fn ensure_liquidity(
+        account: &T::AccountId,
+        reduce_amount: FixedU128,
+        dot_lf: bool,
+    ) -> DispatchResult {
+        let (total_liquidity, _) = Self::get_account_liquidity(account)?;
+        let lf_liquidity = Self::get_account_lf_liquidity(account)?;
```
yrong, line 1866: move this into the if block next line to avoid unnecessary calc

yrong: yes,realize in our current design we need to check lf_liquidity for none-lf-asset. so just ignore this comment

yz89 (approved):

PR #1744, opened 2022-05-11
remove related storage when delete a stream

rjman-ljm:
## Context

Fix a situation by deleting all storage

### Case

Both An and B record stream X. if A need to delete X (reaches the Completed threshold), but B need not delete X, then whether to delete X or not?

### Solution

Delete all

### DIFF
--- a/pallets/streaming/src/lib.rs
+++ b/pallets/streaming/src/lib.rs
@@ -119,8 +119,10 @@ pub mod pallet {
         InsufficientStreamBalance,
         /// Excess max streams count
         ExcessMaxStreamsCount,
+        /// Stream not started
+        NotStarted,
         /// Stream was cancelled or completed
-        StreamHasFinished,
+        HasFinished,
     }
 
     #[pallet::event]
@@ -253,11 +255,11 @@ pub mod pallet {
             );
             Streams::<T>::insert(stream_id, stream);
 
-            // Remove the outdated and finished streams
-            Self::update_finished_stream_library(&sender, &recipient)?;
             // Add the stream_id to stream_library for both the sender and receiver.
             Self::try_push_stream_library(&sender, stream_id, StreamKind::Send)?;
             Self::try_push_stream_library(&recipient, stream_id, StreamKind::Receive)?;
+            // Remove the outdated and finished streams, should do update after push
+            Self::update_finished_stream_library(&sender, &recipient)?;
 
             Self::deposit_event(Event::<T>::StreamCreated(
                 stream_id, sender, recipient, deposit, asset_id, start_time, end_time, true,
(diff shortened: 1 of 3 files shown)


rjman-ljm:
@GopherJ This PR is ready to review

GopherJ (approved):

pallets/streaming/src/lib.rs
```diff
@@ -412,26 +416,29 @@ impl<T: Config> Pallet<T> {
...
             |registry: &mut Option<BoundedVec<StreamId, T::MaxStreamsCount>>| -> DispatchResult {
                 let mut r = registry.take().unwrap_or_default();
-                r.as_mut().sort_unstable_by(|a, b| b.cmp(a));
-
                 let len = r.len() as u32;
                 match len {
-                    _x if len >= T::MaxFinishedStreamsCount::get() => {
+                    _x if len > T::MaxFinishedStreamsCount::get() => {
                         if let Some(stream_id) = r.pop() {
                             if Streams::<T>::get(stream_id).is_some() {
```
GopherJ, line 430: deep nested scope can be improved I guess

rjman-ljm:
```rust
if r.len() as u32 > T::MaxFinishedStreamsCount::get() {
    // This is safe because we have ensured that the length is greater than the limit(at least 0)
    let stream_id = r.pop().unwrap();
    if Streams::<T>::get(stream_id).is_none() {
        return Ok(())
    };
    // This is safe because we have ensured that the stream is in storage
    let stream = Streams::<T>::get(stream_id).unwrap();

    // Remove all related storage
    Self::try_remove_stream_library(&stream.sender, stream_id, None)?;
    Self::try_remove_stream_library(
        &stream.recipient,
        stream_id,
        None,
    )?;
    Streams::<T>::remove(stream_id);
}
*registry = Some(r);
```

rjman-ljm: how about this solution? I've tried it before, but there is a cascading relationship between these logics.

GopherJ: is it possible to quick return?

rjman-ljm: There will be a lot of quick return.

rjman-ljm:
```rust
if r.len() as u32 <= T::MaxFinishedStreamsCount::get() {
    *registry = Some(r);
    return Ok(());
}

// This is safe because we have ensured that the length is greater than the limit(at least 0)
if let Some(stream_id) = r.pop() {
    let stream = Streams::<T>::get(stream_id).ok_or(Error::<T>::InvalidStreamId)?;
    // Remove all related storage
    Self::try_remove_stream_library(&stream.sender, stream_id, None)?;
    Self::try_remove_stream_library(&stream.recipient, stream_id, None)?;
    Streams::<T>::remove(stream_id);
```

PR #1749, opened 2022-05-12
add cancel unstake call

GopherJ:


pallets/liquid-staking/src/types.rs
```diff
@@ -101,19 +101,19 @@ impl<Balance: BalanceT + FixedPointOperand> MatchingLedger<Balance> {
         Ok(())
     }
 
-    fn sub_stake_amount(&mut self, amount: Balance) -> DispatchResult {
+    pub fn sub_stake_amount(&mut self, amount: Balance) -> DispatchResult {
         self.total_stake_amount.total = self
             .total_stake_amount
-            .total
+            .free()?
```
MrPai, line 107: in `consolidate_stake`, `total_stake_amount.total` and `total_stake_amount.reserve` will deduct a same amount, if change to 'free()', then `reserve` has been deduct here, it's better to use `total`.

pallets/liquid-staking/src/lib.rs
```diff
@@ -889,6 +894,56 @@ pub mod pallet {
...
+            Unlockings::<T>::try_mutate(&who, |b| -> DispatchResultWithPostInfo {
+                let chunks = b.as_mut().ok_or(Error::<T>::NoUnlockings)?;
+                let mut cancelled: Balance = Zero::zero();
+
+                while let Some(last) = chunks.last_mut() {
+                    if last.era != Self::current_era() + T::BondingDuration::get() + 1
+                        || cancelled >= amount
+                    {
+                        break;
+                    }
```
MrPai, line 916: This may be always `break`, because `last.era` is calculated by `let target_era = Self::current_era() + T::BondingDuration::get() + 1;` however, this two `Self::current_era()` is different.

pallets/liquid-staking/src/types.rs
GopherJ: yes I think I did a bad change, I'd like to deduct the free part but this seems to be a wrong way

pallets/liquid-staking/src/lib.rs
GopherJ: the current `cancel_unstake` only works when users are cancelling in the same era. Because when we advance era, we will start to go to relaychain to bond

MrPai: this may make users confused if we only allow them cancel_unstake in the same era, there are actually 29 eras for an unlocking period, `cancel_unstake` suppose to let them cancel in any era before the last.
but if they found that it doesn't work during era 2~29 and only works in era1. this doesn't make sense.

GopherJ: yes but feel that should be `fast_unstake` domain if we would like to implement feature like that

PR #1750, opened 2022-05-13
more boundary tests && dealing with the amount of accuray loss

rjman-ljm:
https://github.com/parallel-finance/parallel/pull/1750/files#diff-5762a8635cd245e593744b18252164c4157b8fc14bde3921c1a6cf12c6a96b93L178-R198

The logic has changed here
The balance of recipient should be calculated by:
  - stream has reaches the end_time, return remaining_balance
    -  If we calculated by `delta * per_rate_secs`, we will lost some tiny value
      > because per_rate_sec = deposit_amount / duration, there exist some accuracy loss
      > eg. per_rate_secs = `10100000000000 / 13 = 7769230769230.77 ~ 7769230769230`
      > we cannot withdraw all due to `7769230769230 * 13 = 100999999999990 < 101000000000000`
  - stream has not reach the end_time, needs to calculate by `delta * per_rate_secs`
    - stream has not been withdrawn, return `delta * per_rate_secs`
    - stream has been withdrawn some, return `delta * per_rate_secs) - claimed_balance`

### DIFF
--- a/pallets/streaming/src/lib.rs
+++ b/pallets/streaming/src/lib.rs
@@ -456,8 +456,10 @@ impl<T: Config> Pallet<T> {
         let checked_push =
             |registry: &mut Option<BoundedVec<StreamId, T::MaxStreamsCount>>| -> DispatchResult {
                 let mut r = registry.take().unwrap_or_default();
-                r.try_push(stream_id)
-                    .map_err(|_| Error::<T>::ExcessMaxStreamsCount)?;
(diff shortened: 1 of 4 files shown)


GopherJ (approved):

rjman-ljm:
@GopherJ Can I do these changes? I think updating StreamLibrary shouldn't throw any error.

https://github.com/parallel-finance/parallel/pull/1750/commits/628b7752f84a8ca667e0f30120c4e5cc67699bb0

PR #1762, opened 2022-05-17
allow only supported assets for streaming

rjman-ljm:


pallets/streaming/src/lib.rs
```diff
@@ -51,6 +51,8 @@ pub use pallet::*;
 pub mod weights;
 pub use weights::WeightInfo;
 
+pub const MINIMUM_DEPOSIT_AMOUNT: u128 = 10_000_000_000_000_000_000_000;
```
GopherJ, line 54: I guess this will not work for some tokens with decimals 6. We should only allow to create stream if the minimumDeposit is set I guess

rjman-ljm: Okay, it does make it better

GopherJ (approved):

PR #1768, opened 2022-05-27
Filter lp token with only lf enabled

yrong:
see some chain logs as following, to avoid unnecessary match we filter these lp assets out of the scope

```
2022-05-26 10:46:12.441  WARN ThreadId(95) prices::get_lp_vault_asset_price: Could not get price for asset 6000 will fall back to oracle
2022-05-26 10:46:12.443  WARN ThreadId(95) prices::get_lp_vault_asset_price: Could not get price for asset 6001 will fall back to oracle
2022-05-26 10:46:12.449  WARN ThreadId(95) prices::get_lp_vault_asset_price: Could not get price for asset 6001 will fall back to oracle
2022-05-26 10:46:12.813  WARN ThreadId(95) prices::get_lp_vault_asset_price: Could not get price for asset 6002 will fall back to oracle
```

we can wait to merge this pr after integration test finished

### DIFF
--- a/pallets/prices/src/lib.rs
+++ b/pallets/prices/src/lib.rs
@@ -336,7 +336,7 @@ impl<T: Config> PriceFeeder for Pallet<T> {
                     true => Self::get_staking_asset_detail_price(asset_id, mantissa, base_price),
                     false => match is_vault_token(*asset_id) {
                         true => Self::get_vault_asset_detail_price(asset_id, mantissa, base_price),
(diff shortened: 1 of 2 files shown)


GopherJ:
hi @yrong  for this piece of code, I think we can use:

```
if is_lp_token {
  return ..
}

if is_vault_token {
  return..
}
```

this may look better without many `true` and `false

GopherJ (approved):

PR #1779, opened 2022-05-31
improve cancel_unstake

### LINKED ISSUE #1772: Cancel unstake improvement
currently it supports only cancel at the same era. We should support cancel at any era

MrPai:
close #1772


GopherJ (approved):

GopherJ:
it looks good but will maybe need to think about side effects to other parts

GopherJ (changes requested):
@MrPai I think this approach looks vulnerable because matching_pool's stake amount has been modified but no actual stake amount transferred to pallet account.

This will influence claim_for call

GopherJ:
backlog this feature

PR #1782, opened 2022-06-01
Update unit tests for liquidate borrow

alannotnerd:
Original liquidate borrow doesn't contains lf loans. In this pr, we update related unit tests, as well the `init_setup` to add lf collateral.

Changes:
- Rename dollar to unit, because when price chnaged, dollar(1) doesn't present to $1 token.
- Update unit tests with lf collateral


pallets/loans/src/mock.rs
```diff
@@ -432,7 +432,7 @@ pub fn micro_dollar(d: u128) -> u128 {
 }
 
 pub fn million_dollar(d: u128) -> u128 {
```
GopherJ, line 434: million_unit

GopherJ (approved):

PR #1785, opened 2022-06-04
Feat/mm fast unstake

GopherJ:


pallets/prices/src/lib.rs
```diff
@@ -185,138 +201,71 @@ impl<T: Config> Pallet<T> {
...
+                timestamp: base_price.timestamp,
+            })
     }
 
     fn get_staking_asset_price(
-        asset_id: &CurrencyId,
-        base_price: TimeStampedPrice,
-    ) -> Option<TimeStampedPrice> {
-        Self::try_get_staking_asset_price(asset_id,base_price).or_else(|| {
-            log::warn!(target: "prices::get_staking_asset_price", "Could not get price for asset {:?} will fall back to oracle", asset_id);
```
yrong, line 213: add some log here?

pallets/liquid-staking/src/lib.rs
```diff
@@ -708,13 +737,22 @@ pub mod pallet {
...
+                    let account_borrows =
+                        T::Loans::get_current_borrow_balance(&account_id, staking_currency)?;
+                    T::Loans::do_repay_borrow(
+                        &account_id,
+                        staking_currency,
+                        min(account_borrows, amount),
+                    )?;
+                    let redeem_amount = T::Loans::get_market_info(collateral_currency)?
+                        .collateral_factor
+                        .saturating_reciprocal_mul_ceil(amount);
```
yrong, line 750: should be `min(account_borrows, amount)` here?

GopherJ: it's already `min(account_borrows, amount)` I guess

yrong: https://github.com/parallel-finance/parallel/blob/1a2c4933adf9b1c33356972aaa6c6c97b54d8b06/pallets/liquid-staking/src/lib.rs#L743-L751

`redeem_amount`  here is still calculated base on amount, will it be over redeemed?

GopherJ: I think it shouldn't because `mint_amount` in `do_loans_fast_unstake` is calculated using `liquid_amount * exchangeRate / collateral_factor` but the item in unlockings is `liquid_amount * exchangeRate`

runtime/heiko/src/lib.rs
```diff
@@ -148,7 +148,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_version: 187,
     impl_version: 32,
     apis: RUNTIME_API_VERSIONS,
-    transaction_version: 15,
+    transaction_version: 16,
```
MrPai, line 151: Is there a compatibility issue with the Ledger?

GopherJ: yes but  we changed unstake call's parameters so I think this is a must right

pallets/liquid-staking/src/lib.rs
```diff
@@ -1652,6 +1700,30 @@ pub mod pallet {
             Ok(())
         }
 
+        #[require_transactional]
+        fn do_loans_fast_unstake(who: &AccountIdOf<T>, amount: BalanceOf<T>) -> DispatchResult {
+            let loans_fast_unstake_fee = T::LoansFastUnstakeFee::get()
+                .checked_mul_int(amount)
+                .ok_or(ArithmeticError::Overflow)?;
```
MrPai, line 1707: Is there any connection between `LoansFastUnstakeFee` and `CollateralCurrency.borrow_rate`?
because this affects a lot, whether LiquidStaking pallet can repay_all account_borrows.

pallets/liquid-staking/src/lib.rs
```diff
@@ -708,13 +737,22 @@ pub mod pallet {
...
                 }
 
-                T::Assets::transfer(
-                    Self::staking_currency()?,
-                    &Self::account_id(),
-                    &who,
-                    amount,
-                    false,
-                )?;
+                if who == Self::loans_account_id() {
```
MrPai, line 740: claim_for `loans_account` is so special, maybe we can do it once `advance_era`. or in a hook function.

pallets/liquid-staking/src/lib.rs
GopherJ: @MrPai actually we repay only partially not the whole account_borrows. I'm think maybe we should use `max` instead of `min` to not left KSM/DOT inside liquid staking

pallets/liquid-staking/src/lib.rs
yrong: that is my question also. who will invoke the `do_claim_for` ? from some offchain service or internal onchain hook?

GopherJ: @yrong currently https://github.com/parallel-finance/stake-client will invoke it

MrPai (approved):
Looks good to me, let's test it fully.

PR #1787, opened 2022-06-08
Lf liquidate threshold

alannotnerd:
Allow liquidation threshold different from the collateral factor for lf market

In former, it requires the collateral factor strictly equals to the liquidation threshold due to the calculation.

### DIFF
--- a/pallets/loans/src/lib.rs
+++ b/pallets/loans/src/lib.rs
@@ -1240,6 +1240,7 @@ impl<T: Config> Pallet<T> {
     ) -> Result<(Liquidity, Shortfall, Liquidity, Shortfall), DispatchError> {
         let total_borrow_value = Self::total_borrowed_value(account)?;
         let total_collateral_value = Self::total_liquidation_threshold_value(account)?;
+
         let lf_borrowed_value = Self::get_lf_borrowed_value(account)?;
         let lf_base_position = Self::get_lf_liquidation_base_position(account)?;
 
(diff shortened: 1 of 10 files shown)


pallets/loans/src/lib.rs
```diff
@@ -1121,7 +1121,16 @@ impl<T: Config> Pallet<T> {
         T::PalletId::get().into_account()
     }
 
-    fn get_base_position(account: &T::AccountId) -> Result<FixedU128, DispatchError> {
+    fn get_lf_borrowed_value(account: &T::AccountId) -> Result<FixedU128, DispatchError> {
+        let lf_borrowed_amount =
+            Self::current_borrow_balance(account, T::LiquidationFreeAssetId::get())?;
```
GopherJ, line 1126: we can define a local variable

pallets/loans/src/lib.rs
```diff
@@ -1504,7 +1549,11 @@ impl<T: Config> Pallet<T> {
             repay_amount,
             market
         );
-        let (_, shortfall, _) = Self::get_account_liquidation_threshold_liquidity(borrower)?;
+        let (_, shortfall, lf_liquidity, _) =
+            Self::get_account_liquidation_threshold_liquidity(borrower)?;
+        let shortfall = shortfall
+            .checked_add(&lf_liquidity)
```
GopherJ, line 1555: why it's `add`?

PR #1789, opened 2022-06-11
Add delay for lease end to protect lf collateral from liquidate

yrong:
need further discussion

assume user fail to pay debt in time when lease of parachain matured, their is still chance that his collateral asset in money market(mainly cdot here) could be liquidated, this pr as a workaround to make it less likely to happen


GopherJ:
probably not needed as long as we can update liq-free collaterals in time

PR #1810, opened 2022-06-21
Improve integration test&Move statemine to asset registry

yrong:
- [x] some integration tests to cover xcm transact call and also some refactor accordling
- [x] move statemine asset to asset registry
- [x] more tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9181,6 +9181,7 @@ dependencies = [
  "pallet-prices",
  "pallet-router",
  "pallet-router-rpc-runtime-api",
+ "pallet-staking",
  "pallet-traits",
  "pallet-xcm",
  "pallet-xcm-helper",
--- a/Makefile
+++ b/Makefile
@@ -87,11 +87,16 @@ test-crowdloans:
 
 .PHONY: integration-test
 integration-test:
-	RUST_LOG="xcm=trace,xcm-executor=trace" SKIP_WASM_BUILD= cargo test -p runtime-integration-tests -- --nocapture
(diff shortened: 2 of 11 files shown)


yrong:
@GopherJ  would suggest we start the migration from statemine asset first, if we decide to do we can set storage in heiko reference params here 

https://github.com/parallel-finance/parallel/blob/de75818df4b542784bbb07eaab00c2652e741e16/runtime/heiko/src/lib.rs#L1285-L1296

then in the next upgrade window we have less work to do

GopherJ:
> @GopherJ would suggest we start the migration from statemine asset first, if we decide to do we can set storage into asset-registry first reference params following
> 
> https://github.com/parallel-finance/parallel/blob/de75818df4b542784bbb07eaab00c2652e741e16/runtime/heiko/src/lib.rs#L1285-L1296
> 
> then in the next upgrade window we have less work to do

yes we can start from that

GopherJ:
@yrong can our ksm_per_second function accept also decimals diff of the asset? so that we dont need to hardcode 1e6. we can read asset's deciamls from Assets module or we can also define foreign asset's decimals directly in `constants.rs`

yrong:
> @yrong can our ksm_per_second function accept also decimals diff of the asset? so that we dont need to hardcode 1e6. we can read asset's deciamls from Assets module or we can also define foreign asset's decimals directly in `constants.rs`

@GopherJ  code above is just for hint and will be removed, we need to manually set it into storage as following
image
image


verified in polkadot-9230(v0.9.23 based) + statemine-900(polkadot-v0.9.19 based) which is the running version online
image

usdt could be transfered back and forth with no problem

GopherJ (approved):

PR #1831, opened 2022-08-02
add proxy address

MrPai:


MrPai:
# Test works
## contribute via xcm proxy
image
## contribute via xcm
image

GopherJ (approved):

pallets/crowdloans/src/benchmarking.rs
```diff
@@ -211,6 +211,16 @@ benchmarks! {
         assert_last_event::<T>(Event::VrfUpdated(true).into())
     }
 
+    update_proxy_address {
```
GopherJ, line 214: I think we can make it simple `update_proxy`

pallets/crowdloans/src/benchmarking.rs
```diff
@@ -211,6 +211,16 @@ benchmarks! {
...
+        let caller: T::AccountId = whitelisted_caller();
+    }: _(
+        SystemOrigin::Root,
+        caller.clone()
+    )
+    verify {
+        assert_last_event::<T>(Event::ProxyAddressUpdated(caller).into())
```
GopherJ, line 221: I think we can update this to be simpler `ProxyUpdated`

PR #1835, opened 2022-08-09
Add more hrmp tokens & Fix launching evm image

yrong:
- [x] Improve script for calculating `units_per_second`
- [x] Improve script for hrmp channel
- [x] test integration with ASTAR

### DIFF
--- a/Dockerfile.evm
+++ b/Dockerfile.evm
@@ -1,21 +1,24 @@
 FROM docker.io/paritytech/ci-linux:production as builder
 LABEL description="This is the build stage for Parallel. Here we create the binary."
 
-ARG PROFILE=production
+ARG PROFILE=release
 ARG BIN=parallel
 
 WORKDIR /parallel
 
 COPY . /parallel
 
+RUN rustup default nightly
+
 RUN cargo build --workspace --exclude runtime-integration-tests --profile $PROFILE --bin $BIN --features with-evm-runtime --features runtime-benchmarks --features try-runtime
 
 # ===== SECOND STAGE ======
 
 FROM docker.io/library/ubuntu:20.04
+ENV DEBIAN_FRONTEND=noninteractive
 LABEL description="This is the 2nd stage: a very small image where we copy the Parallel binary."
 
-ARG PROFILE=production
+ARG PROFILE=release
 ARG BIN=parallel
 
 ENV BIN_PATH=/usr/local/bin/$BIN
@@ -23,7 +26,7 @@ ENV BIN_PATH=/usr/local/bin/$BIN
 COPY --from=builder /parallel/target/$PROFILE/$BIN /usr/local/bin
 
 RUN apt update -y \
-    && apt install -y ca-certificates libssl-dev \
+    && apt install -y ca-certificates libssl-dev tzdata \
     && useradd -m -u 1000 -U -s /bin/sh -d /parallel parallel \
     && mkdir -p /parallel/.local \
     && mkdir /data \
--- a/primitives/src/tokens.rs
+++ b/primitives/src/tokens.rs
@@ -44,6 +44,7 @@ pub const IBTC: CurrencyId = 122;
 pub const GENS: CurrencyId = 123;
 pub const EQ: CurrencyId = 124;
 pub const TUR: CurrencyId = 125;
+pub const LIT: CurrencyId = 126;
 
 // Ethereum ecosystem
 pub const EUSDT: CurrencyId = 201;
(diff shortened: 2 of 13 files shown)


yrong:
- Astar & Shiden
https://github.com/AstarNetwork/Astar/blob/e097c3848b2af0d91d6f764125e858d5c9a55add/runtime/astar/src/xcm_config.rs#L52
https://github.com/AstarNetwork/Astar/blob/e097c3848b2af0d91d6f764125e858d5c9a55add/runtime/astar/src/xcm_config.rs#L31

so the mapping `MultiLocation` is just `X1(ParaId)`

primitives/src/tokens.rs
```diff
@@ -44,6 +44,7 @@ pub const IBTC: CurrencyId = 122;
 pub const GENS: CurrencyId = 123;
 pub const EQ: CurrencyId = 124;
 pub const TUR: CurrencyId = 125;
+pub const LIT: CurrencyId = 126;
```
GopherJ, line 47: I think it can be 127 because TUR may has polkadot parachain

scripts/helper/src/utils.ts
```diff
@@ -153,3 +152,33 @@ export const getRelayApi = async (endpoint: string): Promise<ApiPromise> => {
...
+  const weight = 600_000_000
+  //assume we charge 0.02$ at most for each xcm reserved based transfer
+  const max_fee = 0.02
+  /// max_fee = (weight_per_second * weight)/WEIGHT_PER_SECOND/(10**precision) * price
+  /// so weight_per_second = max_fee*WEIGHT_PER_SECOND*(10**precision)/weight/price
+  const weight_per_second = (((max_fee * WEIGHT_PER_SECOND) / weight) * 10 ** precision) / price
+  /// to avoid price sharply increased later so that we charge too much
+  /// just add some soft limit here
+  return Math.min(1000 * WEIGHT_PER_SECOND, Math.floor(weight_per_second))
+}
```
MrPai, line 168: Would like to know if we have any reference documents for this calculation?

Dockerfile.evm
```diff
@@ -1,21 +1,24 @@
...
 
-ARG PROFILE=production
+ARG PROFILE=release
 ARG BIN=parallel
 
 WORKDIR /parallel
 
 COPY . /parallel
 
+RUN rustup default nightly-2022-05-11
```
GopherJ, line 11: we can also ping the `ci-linux:production` to a certain commit I guess

scripts/helper/src/utils.ts
yrong: @MrPai no docs yet but the logic here mainly based on these [xcm configs](https://github.com/parallel-finance/parallel/blob/06be51cb8e9a92fd4f47755917f08434e75727a1/runtime/heiko/src/lib.rs#L1443-L1444)

`Weigher` is to calculate weight consumed for a xcm call and for `polkadotXcm.reservedTransfer` is always `600_000_000`
`Trader` is to calculate how much token equivalent required for these consumed weight and for `FixedRateOfFungible` is just `weight_per_second * weight/WEIGHT_PER_SECOND`

This script is to calculate `weight_per_second` based on that we charge some fixed us dollar(0.02$) for each xcm reserved transfer, so we add token price and precision as param here, we can get token price from some public exchange.

The only tweak I made here is considering currently price for tokens in polkadot ecosystem are somewhat too low, just add some soft limit here to avoid charge too much if price increased later when market gets better.

Dockerfile.evm
```diff
@@ -1,29 +1,32 @@
...
 
-ARG PROFILE=production
+ARG PROFILE=release
 ARG BIN=parallel
 
 WORKDIR /parallel
 
 COPY . /parallel
 
+RUN rustup default nightly
```
GopherJ, line 11: I feel this will still break the docker image build in the future, rust nightly is very unstable. Are we able to use `rust-toolchain` or use `ci-linux:<commit>`

GopherJ (approved):

Dockerfile.evm
yrong: nightly here means toolchain from docker.io/paritytech/ci-linux:production and now is 2022-07-25 https://github.com/paritytech/scripts/blob/1c0fde6a6eb6b1a700da7261b3b6be0463d26d17/dockerfiles/ci-linux/Dockerfile#L24

this should be tested by parity team and I think always sync with them should be fine. If anything breaking in our code not compatible with it we can just make some fix

PR #1841, opened 2022-08-12
Add price for lc_dot & Relocate evm dependencies

yrong:


pallets/prices/src/lib.rs
```diff
@@ -154,6 +160,19 @@ pub mod pallet {
...
+
+        /// Set foreign vault token mapping
+        #[pallet::weight((<T as Config>::WeightInfo::set_foreign_vault_mapping(), DispatchClass::Operational))]
+        #[transactional]
+        pub fn set_foreign_vault_mapping(
+            origin: OriginFor<T>,
+            foreign_asset_id: CurrencyId,
+            asset_id: CurrencyId,
+        ) -> DispatchResultWithPostInfo {
+            T::FeederOrigin::ensure_origin(origin)?;
```
GopherJ, line 172: The FeederOrigin naming looks wierd in this case

pallets/prices/src/lib.rs
```diff
@@ -154,6 +160,19 @@ pub mod pallet {
             <Pallet<T> as EmergencyPriceFeeder<CurrencyId, Price>>::reset_emergency_price(asset_id);
             Ok(().into())
         }
+
+        /// Set foreign vault token mapping
+        #[pallet::weight((<T as Config>::WeightInfo::set_foreign_vault_mapping(), DispatchClass::Operational))]
+        #[transactional]
+        pub fn set_foreign_vault_mapping(
```
GopherJ, line 167: if we are doing  mapping from assetId to assetId then it can be more general and has nothing to do with `vault` anymore.

if we are using `vault` wording, I think then we can map from assetId to `leaseStart, leaseEnd`

pallets/prices/src/lib.rs
yrong: how about just use sudo here

PR #1842, opened 2022-08-13
Fix redeem issue

alannotnerd:


pallets/loans/src/lib.rs
```diff
@@ -1429,7 +1431,7 @@ impl<T: Config> Pallet<T> {
         })?;
 
         T::Assets::transfer(asset_id, &Self::account_id(), who, redeem_amount, false)
-            .map_err(|_| Error::<T>::InsufficientMarketLiquidity)?;
+            .map_err(|_| Error::<T>::InsufficientCash)?;
```
GopherJ, line 1434: do we still need `InsufficientMarketLiquidity`

pallets/loans/src/tests.rs
```diff
@@ -274,7 +271,7 @@ fn redeem_fails() {
 }
 
 #[test]
-fn withdraw_fails_when_insufficient_liquidity() {
+fn redeem_fails_when_insufficient_liquidity() {
```
GopherJ, line 274: should be in another unit tests, this is for testing normal redeem

alannotnerd: done

pallets/loans/src/lib.rs
alannotnerd: I think `InsufficientMarketLiquidity` should only appear in `ensure_liquidity`. Alternatively, We can add a new error for `PalletAccountBalanceLow` for this situation.

PR #1843, opened 2022-08-15
add MatchingPool fast unstake logic

### LINKED ISSUE #1834: LiquidStaking: fast unstake based on matching pool
Currently, instant unstake fee based on `Loans` charged high, now design instant unstake based on matching pool charged less, and not instantly. called it `fast unstake`

MrPai:
close #1834

### DIFF
--- a/pallets/liquid-staking/src/lib.rs
+++ b/pallets/liquid-staking/src/lib.rs
@@ -279,8 +279,8 @@ pub mod pallet {
         /// Commission rate was updated
         CommissionRateUpdated(Rate),
         /// Fast Unstake Matched
-        /// [unstaker, received_staking_amount, liquid_amount,]
-        FastUnstakeMatched(T::AccountId, BalanceOf<T>, BalanceOf<T>),
+        /// [unstaker, received_staking_amount, matched_liquid_amount, fee_in_liquid_currency]
+        FastUnstakeMatched(T::AccountId, BalanceOf<T>, BalanceOf<T>, BalanceOf<T>),
     }
 
     #[pallet::error]
@@ -1717,15 +1717,13 @@ pub mod pallet {
 
         #[require_transactional]
         pub fn do_matching() -> DispatchResult {
-            let remove_results = FastUnstakeRequests::<T>::clear(200u32, None);
             let total_unbonding = Self::get_total_unbonding();
             let (bond_amount, rebond_amount, unbond_amount) =
                 Self::matching_pool().matching(total_unbonding)?;
 
             log::trace!(
                 target: "liquidStaking::do_matching",
-                "remove_results: {:?}, bond_amount: {:?}, rebond_amount: {:?}, unbond_amount: {:?}",
-                remove_results.loops,
+                "bond_amount: {:?}, rebond_amount: {:?}, unbond_amount: {:?}",
                 &bond_amount,
                 &rebond_amount,
                 &unbond_amount
@@ -1870,39 +1868,53 @@ pub mod pallet {
                         .ok_or(Error::<T>::InvalidExchangeRate)?;
 
                 let matched_liquid_amount = request_liquid_amount.min(available_liquid_amount);
-                if matched_liquid_amount.is_zero() {
-                    return Ok(());
+
+                if !matched_liquid_amount.is_zero() {
+                    let matched_fee = T::MatchingPoolFastUnstakeFee::get()
+                        .saturating_mul_int(matched_liquid_amount);
+                    let liquid_to_burn = matched_liquid_amount.saturating_sub(matched_fee);
+                    T::Assets::burn_from(Self::liquid_currency()?, unstaker, liquid_to_burn)?;
+                    T::Assets::transfer(
+                        Self::liquid_currency()?,
+                        unstaker,
+                        &T::ProtocolFeeReceiver::get(),
+                        matched_fee,
+                        false,
+                    )?;
+
+                    let staking_to_receive = Self::liquid_to_staking(liquid_to_burn)
+                        .ok_or(Error::<T>::InvalidExchangeRate)?;
+
+                    MatchingPool::<T>::try_mutate(|p| p.sub_stake_amount(staking_to_receive))?;
+                    T::Assets::transfer(
+                        Self::staking_currency()?,
+                        &Self::account_id(),
+                        unstaker,
+                        staking_to_receive,
+                        false,
+                    )?;
+
+                    Self::deposit_event(Event::<T>::FastUnstakeMatched(
+                        unstaker.clone(),
+                        staking_to_receive,
+                        matched_liquid_amount,
+                        matched_fee,
+                    ));
                 }
-                T::Assets::burn_from(Self::liquid_currency()?, unstaker, matched_liquid_amount)?;
 
-                let matched_staking_amount = Self::liquid_to_staking(matched_liquid_amount)
-                    .ok_or(Error::<T>::InvalidExchangeRate)?;
-
-                let receive_amount = Rate::one()
-                    .saturating_sub(T::MatchingPoolFastUnstakeFee::get())
-                    .saturating_mul_int(matched_staking_amount);
-
-                // bond fee to relaychain, distribute rewards to all users
-                MatchingPool::<T>::try_mutate(|p| p.sub_stake_amount(receive_amount))?;
-
-                T::Assets::transfer(
-                    Self::staking_currency()?,
-                    &Self::account_id(),
-                    unstaker,
-                    receive_amount,
-                    false,
-                )?;
-
-                // decrease storage `FastUnstakeRequests`
                 let unmatched_amount = request_liquid_amount.saturating_sub(matched_liquid_amount);
                 if !unmatched_amount.is_zero() {
                     *b = Some(unmatched_amount);
                 }
-                Self::deposit_event(Event::<T>::FastUnstakeMatched(
-                    unstaker.clone(),
-                    receive_amount,
-                    matched_liquid_amount,
-                ));
+
+                log::trace!(
+                    target: "liquidStaking::do_fast_match_unstake",
+                    "unstaker: {:?}, request_liquid_amount: {:?}, unmatched_amount: {:?}",
+                    unstaker,
+                    request_liquid_amount,
+                    unmatched_amount,
+                );
+
                 Ok(())
             })
         }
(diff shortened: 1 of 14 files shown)


pallets/liquid-staking/src/lib.rs
```diff
@@ -1732,6 +1765,75 @@ pub mod pallet {
             Ok(())
         }
 
+        #[require_transactional]
+        fn do_fast_match_unstake(unstaker: &T::AccountId) -> DispatchResult {
+            FastUnstakeRequests::<T>::try_mutate_exists(unstaker, |b| -> DispatchResult {
+                if let Some(request_amount) = b.take() {
+                    let available_amount = Self::matching_pool().total_stake_amount.free()?;
+                    let actual_amount = available_amount.min(request_amount);
```
GopherJ, line 1773: so we are allowing partially matched right

MrPai: Yes

pallets/liquid-staking/src/lib.rs
```diff
@@ -1627,13 +1658,15 @@ pub mod pallet {
 
         #[require_transactional]
         pub fn do_matching() -> DispatchResult {
+            let remove_results = FastUnstakeRequests::<T>::clear(u32::max_value(), None);
```
yrong, line 1661: may not be safe to clear all here, better to add some limit(e.g:200)  and re-benchmark [this call](https://github.com/parallel-finance/parallel/blob/0d2a2995b2ab2b9f1e75fd98c39d8d3e47410155/pallets/liquid-staking/src/benchmarking.rs#L345-L361) based on the worst case.
Another option is to remove this storage and just add a flag(allow fast unstake or not) in `Unlockings` like what acala implement now

pallets/liquid-staking/src/lib.rs
```diff
@@ -962,6 +979,20 @@ pub mod pallet {
                 Ok(().into())
             })
         }
+
+        /// Fast match unstake through matching pool
+        #[pallet::weight(<T as Config>::WeightInfo::fast_match_unstake(unstaker_list.len() as u32))]
+        #[transactional]
+        pub fn fast_match_unstake(
+            origin: OriginFor<T>,
+            unstaker_list: Vec<T::AccountId>,
```
GopherJ, line 988: should be bounded limited but we can test what will happen

pallets/liquid-staking/src/lib.rs
```diff
@@ -1732,6 +1765,75 @@ pub mod pallet {
             Ok(())
         }
 
+        #[require_transactional]
+        fn do_fast_match_unstake(unstaker: &T::AccountId) -> DispatchResult {
+            FastUnstakeRequests::<T>::try_mutate_exists(unstaker, |b| -> DispatchResult {
+                if let Some(request_amount) = b.take() {
```
GopherJ, line 1771: we can quick return

pallets/liquid-staking/src/lib.rs
```diff
@@ -1732,6 +1765,75 @@ pub mod pallet {
...
+                    let available_amount = Self::matching_pool().total_stake_amount.free()?;
+                    let actual_amount = available_amount.min(request_amount);
+
+                    if !actual_amount.is_zero() {
```
GopherJ, line 1775: we can maybe quick return

pallets/liquid-staking/src/lib.rs
```diff
@@ -1732,6 +1765,75 @@ pub mod pallet {
...
+                        let fee = T::MatchingPoolFastUnstakeFee::get()
```
GopherJ, line 1776: need to put fee into `TotalReserves`

pallets/liquid-staking/src/lib.rs
```diff
@@ -1732,6 +1765,75 @@ pub mod pallet {
...
+                            .checked_mul_int(actual_amount)
+                            .ok_or(ArithmeticError::Overflow)?;
+                        let transfer_amount = actual_amount
+                            .checked_sub(fee)
+                            .ok_or(ArithmeticError::Underflow)?;
+
+                        // decrease storage `Unlockings`
+                        Unlockings::<T>::try_mutate(unstaker, |b| -> DispatchResult {
```
GopherJ, line 1784: need to reuse rebond function

pallets/liquid-staking/src/lib.rs
```diff
@@ -1732,6 +1765,75 @@ pub mod pallet {
...
+            FastUnstakeRequests::<T>::try_mutate_exists(unstaker, |b| -> DispatchResult {
+                if let Some(request_amount) = b.take() {
+                    let available_amount = Self::matching_pool().total_stake_amount.free()?;
+                    let actual_amount = available_amount.min(request_amount);
+
+                    if !actual_amount.is_zero() {
+                        let fee = T::MatchingPoolFastUnstakeFee::get()
+                            .checked_mul_int(actual_amount)
+                            .ok_or(ArithmeticError::Overflow)?;
+                        let transfer_amount = actual_amount
```
GopherJ, line 1779: we have too many amounts util here, I suggest that we limit to:

- amount : unstake_amount * exchangeRate
- matching_fees : fee amount
- receive_amount: `(amount - matching_fees).min(available_amount)`
- matched

pallets/liquid-staking/src/lib.rs
GopherJ: we can check substrate staking pallet on how do they clear

pallets/liquid-staking/src/lib.rs
```diff
@@ -1732,6 +1765,75 @@ pub mod pallet {
...
+                                    break;
+                                }
+                            }
+
+                            Ok(())
+                        })?;
+
+                        // decrease storage `MatchingPool`
+                        MatchingPool::<T>::try_mutate(|p| -> DispatchResult {
+                            p.sub_stake_amount(actual_amount)?;
```
GopherJ, line 1808: we can support another mode which is not matching against the pool but matching against the `fast_match_unstake` caller.

this means:

- we dont `sub_stake_amount` from the pool
- we substract the caller's balance instead
- `fast_match_unstake` will need to add another parameter `stake_amount`

this looks like `matcher` can also stake at the same time with some extra privileges for them to do this. Like:

- we probably dont need to charge staking_fee anymore because this is just sDOT exchange, no new sDOT minted thus dont share rewards

GopherJ: @MrPai how do you think about this? 

Though we can finish first the other parts and then review this again later

MrPai: This could be an optional way, and it should be decided by FE how to do the match.

pallets/liquid-staking/src/lib.rs
MrPai: @yrong one concern for another option is: what if one user both do unstake and fast_unstake in the same era, then we can not distinguish the amount from `Unlockings`.
Acala has two storage: `RedeemRequests ` and `Unbondings `, once the era end, they will move all from `RedeemRequests` to `Unbondings`.

MrPai: @GopherJ I think the underlying code of remove_prefix also has a `limit` argument:
https://paritytech.github.io/substrate/master/frame_support/pallet_prelude/struct.StorageDoubleMap.html#method.remove_prefix
"if limit is none then all values in the client backend are deleted."

yrong: > 

@MrPai then instead of clearing all `FastUnstakeRequests` when era advanced how about we also iter the storage to `do_fast_match_unstake` to match these request first before `do_matching`

pallets/liquid-staking/src/lib.rs
```diff
@@ -1813,6 +1852,63 @@ pub mod pallet {
             Ok(inflate_liquid_amount)
         }
 
+        #[require_transactional]
+        fn do_fast_match_unstake(unstaker: &T::AccountId) -> DispatchResult {
+            FastUnstakeRequests::<T>::try_mutate_exists(unstaker, |b| -> DispatchResult {
+                if b.is_none() {
```
GopherJ, line 1858: maybe ok_or is cleaner than `return + take().expect`

pallets/liquid-staking/src/lib.rs
```diff
@@ -1813,6 +1852,63 @@ pub mod pallet {
...
+                    return Ok(());
+                }
+                let request_liquid_amount =
+                    b.take()
+                        .expect("Could not be none, qed;")
+                        .min(T::Assets::reducible_balance(
+                            Self::liquid_currency()?,
+                            unstaker,
```
GopherJ, line 1866: need to define a variable for:


```
T::Assets::reducible_balance(
                            Self::liquid_currency()?,
                            unstaker,
                            false,
                        )
```

to make the code sounds more clear

pallets/liquid-staking/src/lib.rs
```diff
@@ -1813,6 +1852,63 @@ pub mod pallet {
...
+                            false,
+                        ));
+
+                let available_liquid_amount =
+                    Self::staking_to_liquid(Self::matching_pool().total_stake_amount.free()?)
```
GopherJ, line 1871: we probably dont need this conversion

pallets/liquid-staking/src/lib.rs
MrPai: use `take()` then the value becomes `None`, and it can be deleted if we don't give a new value.

pallets/liquid-staking/src/lib.rs
MrPai: we need to convert liquid_amount to staking_amount or convert vice versa, so they can compare with each other.

pallets/liquid-staking/src/lib.rs
```diff
@@ -1680,13 +1717,15 @@ pub mod pallet {
 
         #[require_transactional]
         pub fn do_matching() -> DispatchResult {
+            let remove_results = FastUnstakeRequests::<T>::clear(200u32, None);
```
yrong, line 1720: if clear with limit here, better to use [CountedStorageMap](https://paritytech.github.io/substrate/master/frame_support/storage/types/struct.CountedStorageMap.html)

yrong: And also better to re-benchmark  the weight of `force_matching` based on the worst case here

GopherJ: though with limit we will probably need to clear multiple times, in this case shall we set a limit for one era's unstake users?

MrPai: what I am thinking is if we use `StorageMap`, then we can store as many requests as we can in the current era, clear 200 is due to the weight limit.
In the current process, will check liquid balance when doing fast matches, so no need to clear all requests.
users' requests in the previous era can still be matched in the next era.

PR #1864, opened 2022-09-01
Feat: distribute crowdloan rewards

MrPai:
- [x] add to runtime
- [x] add tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6911,6 +6911,7 @@ dependencies = [
  "pallet-balances",
  "pallet-currency-adapter",
  "pallet-timestamp",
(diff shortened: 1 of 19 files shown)


MrPai:
There are still some items to be decided:
- [ ] how to calculate the reward amount, if the reward is the same according to the lease period or according to the project?
- [ ] how to update stream start time and end time in crowdloan, store it in `StorageValue` and update it by governance? 
- [ ] if all the project in one lease period has the same stream period(include start time and end time)?

GopherJ:
LGTM!

GopherJ (approved):

PR #1866, opened 2022-09-05
Fix/claimed balance under ed in stream

MrPai:


pallets/streaming/src/lib.rs
```diff
@@ -349,6 +353,10 @@ pub mod pallet {
                 Error::<T>::InsufficientStreamBalance
             );
 
+            if amount.saturating_add(T::ExistentialDeposit::get()) >= recipient_balance {
```
GopherJ, line 356: stream supports KSM/DOT as well. So this line may shouldn't be applied to them

MrPai: So if it's better to use `StorageMap` to store this, and update it by governance?

GopherJ (approved):

PR #1874, opened 2022-09-14
Fix xcm transfer

yrong:
Fix polkadot xcm transfer and add more tests

### DIFF
--- a/integration-tests/src/setup.rs
+++ b/integration-tests/src/setup.rs
@@ -37,7 +37,9 @@ pub const DOT_DECIMAL: u32 = 10;
 pub const RMRK_DECIMAL: u8 = 10;
 pub const USDT_DECIMAL: u8 = 6;
 pub const HEIKO_DECIMAL: u8 = 12;
+pub const PARA_DECIMAL: u8 = 12;
 pub const KAR_DECIMAL: u8 = 12;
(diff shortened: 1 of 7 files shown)


integration-tests/src/setup.rs
```diff
@@ -392,6 +408,44 @@ impl ExtBuilder {
...
+            .unwrap();
+            let clv_asset_location = MultiLocation::new(
+                1,
+                X2(
+                    Parachain(2002),
+                    //since we use para to mock clv,just use para location here
+                    GeneralKey(WeakBoundedVec::<u8, ConstU32<32>>::force_from(
+                        b"PARA".to_vec(),
+                        None,
+                    )),
```
MrPai, line 438: Looks like if using `b"PARA"` here, then this line will take effects:
https://github.com/parallel-finance/parallel/blob/1a93d800a816ae307ea0c6e3a32e2f30d15f6345/runtime/parallel/src/lib.rs#L492-L497

`AssetRegistry` will not take any effect here.

GopherJ (approved):

integration-tests/src/setup.rs
GopherJ: yes it's only used as fallback

PR #1901, opened 2022-10-28
Fix: check bonus config in crowdloan

MrPai:
Add more checks when updateBonusconfig.


MrPai:
@GopherJ pls check this PR, and do you think if we need to ensure `start_time>current_time` when updating bonus config, like mentioned in this card: https://parallelfi.atlassian.net/browse/BTE-962

GopherJ (approved):

PR #1957, opened 2023-12-26
fix: min nominator bond

0x8f701:


MrPai (approved):
LGTM, can be merged after fix the CI.

